PR #25, opened 2025-12-02
[filler-words] - remove filler words

EliseiNicolae:
Description: 
Removed filler words like `'um' / 'uh', 'eh'` etc + added editor to be able to add more.

Screenshot 2025-12-02 at 10 05 16

---
@altic-dev - btw I think the email: altic_dev@gmail.com hasn't been created because i got:
image

### DIFF
--- a/Sources/Fluid/ContentView.swift
+++ b/Sources/Fluid/ContentView.swift
@@ -981,11 +981,40 @@ struct ContentView: View {
                             }
                             .font(.caption)
                             .foregroundStyle(.secondary)
+
+                            Divider()
+                                .padding(.vertical, 4)
+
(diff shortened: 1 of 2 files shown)


altic-dev:
Hey! Thanks for the contribution. Where did you get my email from btw? alticdev@gmail.com works.  Sorry about that.

This is a nice feature btw! Love it. But I think it would suit better in the AI setting area right below the model selection so it makes sense for people to change it there. Preferences imo is for actual app settings and other configurations and not related to model or post processing. What do you think?

EliseiNicolae:
Got the email from website:
https://www.altic.dev/contact

image

PR #26, opened 2025-12-02
[update-ui] - update the listening container to be at the bottom

EliseiNicolae:
UI:

Screenshot 2025-12-02 at 10 48 13

### DIFF
--- a/Sources/Fluid/Services/NotchOverlayManager.swift
+++ b/Sources/Fluid/Services/NotchOverlayManager.swift
@@ -21,56 +21,56 @@ enum OverlayMode: String {
 @MainActor
 final class NotchOverlayManager {
     static let shared = NotchOverlayManager()
-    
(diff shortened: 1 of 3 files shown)


altic-dev:
Okay this one is debatable because people seem to love the notch. My earlier versions used to have it below like your screenshot. Perhaps an option to choose between 2 makes sense but not sure. Let's see if we get more responses on this to see what to end up with. Why don't you prefer the notch, if I may ask?

EliseiNicolae:
That's ok, i think we can close it.
The notch on a big monitor 31 inch, it's a little bit incomod because i have to check every time if is recording or not by moving my head :)), and that's why I'm keeping it at the bottom.

altic-dev:
will work on it to bring it to the bottom as well as optional :) Will use the code from this PR or if you want to do it as well, I am all in for that as well!

PR #76, opened 2026-01-03
Fix integration with local Ollama models

jiwaszki:
## Description

> **I want to open the discussion on the topic of running local Ollama models with FluidVoice.**

> Short story: I have installed latest Ollama and downloaded few models trying to hook them into this great app. I have faced few issues, thus this PR was made.

This PR fixes three issues:
1. Allow local Ollama's models to run by adjusting json parsing to current standards of API. (ref: https://github.com/ollama/ollama/blob/main/docs/api.md#generate-a-chat-completion)

2. Fix passing of `config.streaming` value -- Ollama models require `stream` value to be explicitly set to `false`, otherwise they default to `true`.

3. Problem with cases where streaming parser solution cannot accurately detect the switch between "think"/"content" switch:
    ```
    Stream: {"thinking": "reasoning...", "content": ":"} → {"thinking": "more...", "content": ":"} → {"content": "Subject:"}
    ```
    I have encountered this problem as all models (tested by me) use `StandardThinkingParser` as the default. I know it is already mentioned that `we let the parser decide unless it's a known separate-field model`. But this case is a little bit different, it includes both fields and only removes "thinking" later on. I have spotted this behaviour on Qwen-based models.

Also I am attaching the screenshot of my setting that I have used to test the solution (with Enable Streaming on and off). Additionally, I have used Apple Intelligence to verify correctness.

## Type of Change
- [ ] 📝 Documentation update

## Testing
- [ ] Tested on Intel/Apple Silicon Mac
- [x] Tested on macOS 26.2
- [ ] Ran linter locally: `brew install swiftlint && swiftlint --strict --config .swiftlint.yml`
- [ ] Ran formatter locally: `brew install swiftformat && swiftformat --config .swiftformat Sources`

## Screenshots / Video 
Image 03-01-2026 at 23 33

## Summary by CodeRabbit

* **New Features**
  * Added automatic runtime detection for AI models that use separate thinking and reasoning fields
  * Expanded support for diverse response formats from different AI service providers and local models

* **Bug Fixes**
  * Improved streaming response parsing to accommodate varied message data formats
  * Enhanced streaming request handling for more consistent and reliable operation

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/Sources/Fluid/Services/LLMClient.swift
+++ b/Sources/Fluid/Services/LLMClient.swift
@@ -191,9 +191,7 @@ final class LLMClient {
         }
 
         // Add streaming flag
-        if config.streaming {
-            body["stream"] = true
-        }
+        body["stream"] = config.streaming
 
         // Add extra parameters in layers:
         // 1. Model-specific parameters (from ThinkingParserFactory)
@@ -262,9 +260,7 @@ final class LLMClient {
         DebugLogger.shared.debug("LLMClient: Non-streaming response received (\(data.count) bytes)", source: "LLMClient")
 
         guard let json = try JSONSerialization.jsonObject(with: data) as? [String: Any],
-              let choices = json["choices"] as? [[String: Any]],
-              let choice = choices.first,
-              let message = choice["message"] as? [String: Any]
+              let message = json["message"] as? [String: Any]
         else {
             throw LLMError.invalidResponse
         }
@@ -296,6 +292,9 @@ final class LLMClient {
         var contentBuffer: [String] = []
         var tagDetectionBuffer = ""
 
+        // Track if we've seen separate thinking fields (for runtime detection)
+        var hasSeenSeparateThinkingField = false
+
         // Tool call accumulation
         var toolCallId: String?
         var toolCallName: String?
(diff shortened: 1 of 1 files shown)


altic-dev:
Is reasoning models the only issue or does it not work with Non reasoning/ instruct models too?

altic-dev:
No no, your claim is valid ofc! We can definitely add it as a replacement for current ollama. I don't care if its chat completions or not. But I still cannot understand why it doesn't take in thinking params well because everything else does it so well. Qwen specifically does not support thinking option itself. So your ollama way i think writes a proxy for 'think' to work.

For qwen : you need to add '/no_think' to the prompt itself and it wont think. That's how the model is built. It does not take a 'think' : false. Try adding that to prompt and see how it works. Also, there's something called 'instruct' version of the qwen. 

Give the above one a try and see what happens :)) 

PS: just an FYI, model weights sont increase if there's thinking. The extra space is stored for something called KV caching. It caches your input/output for faster processing :) and you can set a limit for it!

altic-dev:
clsing as it's inactive. please open up a new one after trying out my suggestions or anything else ofc :)

PR #145, opened 2026-02-07
Fix app freeze when USB microphone is plugged in during recording

JereSalo:
## Problem

When a USB microphone (in my case, a HyperX SoloCast) is plugged in while recording — or plugged in and recording is started immediately after — the entire app freezes. The UI becomes completely unresponsive and must be force-quit. Any in-progress transcription is lost.

The freeze is not a crash (no crash logs) — it's a deadlock. CoreAudio property change notifications fire while CoreAudio holds an internal lock, and the listener callbacks immediately call back into CoreAudio (`listInputDevices()`, `getDefaultInputDevice()`, etc.) on the main thread, waiting for the same lock indefinitely.

### 1. Prevent the freeze on audio device changes

- Defer all CoreAudio listener callbacks to the next runloop pass (`DispatchQueue.main.async`) so they run outside the callback context where the lock is held.
- Move CoreAudio device queries in `handleDeviceListChanged()` and `refreshDevices()` to a background thread.
- Release the old `AVAudioEngine` on a background thread in `stopWithoutTranscription()` — deallocating an engine whose underlying device just died can also block the main thread.

### 2. Keep recording alive through device changes

Previously, if the audio device died during recording (which happens when plugging in a USB mic — macOS destroys and recreates the aggregate device), the app cancelled the recording.

Now it hot-swaps to a fallback device instead:
1. Tears down the dead engine
2. Waits 300ms for CoreAudio to finish the topology change
3. Queries available devices on a background thread
4. Rebuilds the engine with the new system default (or first available input)
5. Continues recording seamlessly

If no fallback device is available, the recording is cancelled as before.

### DIFF
--- a/Sources/Fluid/Services/ASRService.swift
+++ b/Sources/Fluid/Services/ASRService.swift
@@ -1703,10 +1703,6 @@ final class ASRService: ObservableObject {
     // Device caching for change detection
     private var cachedDeviceUIDs: Set<String> = []
 
-    private func checkDeviceWasAvailable(_ uid: String) -> Bool {
-        return self.cachedDeviceUIDs.contains(uid)
-    }
-
     private func cacheCurrentDeviceList(_ devices: [AudioDevice.Device]) {
         self.cachedDeviceUIDs = Set(devices.map { $0.uid })
     }


JereSalo:
I've been running a build from this branch as my daily driver. If I notice any issues or unexpected behavior I'll report back here.

altic-dev:
Perfect! Keep using it and we can merge it in a week's time if it all works well :)) thanks for helping!

PR #187, opened 2026-03-04
feat(linux): initial Linux support with NVIDIA Parakeet

mpnikhil:
Introduces initial support for Linux including:
- `linux/app.py`: Core transcription application using NVIDIA Parakeet via NeMo.
- `linux/install_and_run.sh`: Robust installer with system dependency checks and model weight relocation.
- `linux/start.sh` & `linux/stop.sh`: Background process management.
- Updated `.gitignore` to exclude the `install/` directory.

The application captures audio while [Alt] is held and transcribes directly into the focused window. Generic paths are used throughout to support different user environments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -208,6 +208,7 @@ create_dmg_only.sh
 # Scratch directory
 scratch/
 
+install/
 assets/
 .swiftpm/
 scripts/
--- a/linux/install_and_run.sh
+++ b/linux/install_and_run.sh
@@ -1,63 +1,88 @@
 #!/bin/bash
-set -e
 
-# Default storage location
-DEFAULT_STORAGE="$HOME/.local/share/fluidvoice"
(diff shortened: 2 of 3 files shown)


mpnikhil:
> Hey @mpnikhil thanks for the PR, I have some general concerns. This seems like a barebones implementation of running inference using the parakeet model. If we are thinking about linux, I would think about windows too in order to be cross platform. I would prefer to build an MVP using Tauri or Electron that can run on both platforms, not just a python script we start and stop using shell scripts. Let me know if you have any thoughts are ideas around this. We can discuss

Fair. Although my push is towards explicit minimalism. Little to no UI (except for the recording overlay).  Fire and forget for p99 usecase. Let me know what you think. This works and is very fast. Linux users should be comfortable using scripts. Any suggestions for improving WER?

PR #190, opened 2026-03-05
perf(asr): reduce recording start latency from ~650ms to ~37ms

floze-the-genius:
## Summary

Reduces the time between pressing the hotkey and audio capture actually starting from **~650-1030ms to ~37-40ms** by optimizing every stage of the pipeline.

## Changes by stage

| Stage | Before | After | File |
|-------|--------|-------|------|
| Hold-mode delay | 150ms | 0ms | `GlobalHotkeyManager.swift` |
| Task dispatch gap | 234-415ms | 0ms (direct await) | `ContentView.swift` |
| Engine recreation | 350-580ms | 0ms (retained) | `ASRService.swift` |
| Media pause | blocking XPC | concurrent | `ASRService.swift` |
| Overlay UI setup | before ASR | after ASR | `ContentView.swift` |
| ensureAsrReady() | always ~67ms | skipped if loaded | `ASRService.swift` |
| Analytics/history | blocking | background Task | `ContentView.swift` |
| Focus-settle sleep | 80ms | 30ms | `ContentView.swift` |

### Key changes

**`GlobalHotkeyManager.swift`**: Removed 150ms hold-mode detection delay. Recording starts immediately on modifier keydown instead of waiting to rule out key combos.

**`ContentView.swift`**: Made `startRecording()` async with direct `await asr.start()` instead of `Task { await asr.start() }`. The Task wrapper enqueued the start on MainActor *after* SwiftUI layout updates triggered by `setActiveRecordingMode()`, causing a 234-415ms dispatch gap. Also moved `setOverlayMode()` calls to run after recording starts (overlay UI setup was blocking for ~1s of SwiftUI rendering). Analytics and history saves moved to background Tasks.

**`ASRService.swift`**: 
- **Engine retention**: Instead of destroying `engineStorage` after every `stop()` and recreating on `start()`, the engine is now retained in stopped state. This avoids ~350-580ms of CoreAudio aggregate device setup. Device-change listeners still trigger recreation when needed.
- **Immediate `isRunning`**: Set to `true` right after engine start, before media pause. Media pause runs concurrently in a background Task.
- **Non-blocking streaming cancel**: `stop()` cancels the streaming task without awaiting completion, since the buffer copy is thread-safe.
- **Skip redundant readiness checks**: `ensureAsrReady()` and `configureSession()` are skipped when the model/engine are already loaded.

## Risk assessment

- Engine retention changes object lifecycle — tested across repeated start/stop cycles, language switches, device hot-plug, and error recovery paths
- Concurrent media pause benefits from #189 (`ContinuationOnceGuard`) for safety under edge cases
- Zero-delay hold-mode means key combos involving the hotkey modifier may trigger a brief recording start before cancellation — acceptable tradeoff for the latency improvement

## Testing

Measured with debug log timestamps across 20+ start/stop cycles. Consistent ~37-40ms from keydown to `isRunning = true`.

### DIFF
--- a/Sources/Fluid/ContentView.swift
+++ b/Sources/Fluid/ContentView.swift
@@ -928,7 +928,7 @@ struct ContentView: View {
             isTranscriptionFocused: self.$isTranscriptionFocused,
             accessibilityEnabled: self.accessibilityEnabled,
             stopAndProcessTranscription: { await self.stopAndProcessTranscription() },
-            startRecording: self.startRecording,
+            startRecording: { Task { await self.startRecording() } },
             openAccessibilitySettings: self.openAccessibilitySettings
         )
     }
@@ -1063,7 +1063,7 @@ struct ContentView: View {
             copyToClipboard: self.$copyToClipboard,
             hotkeyManager: self.hotkeyManager,
             menuBarManager: self.menuBarManager,
-            startRecording: self.startRecording,
+            startRecording: { Task { await self.startRecording() } },
             refreshDevices: self.refreshDevices,
             openAccessibilitySettings: self.openAccessibilitySettings,
             restartApp: self.restartApp,
@@ -1076,7 +1076,7 @@ struct ContentView: View {
         RecordingView(
             appear: self.$appear,
             stopAndProcessTranscription: { await self.stopAndProcessTranscription() },
-            startRecording: self.startRecording
+            startRecording: { Task { await self.startRecording() } }
         )
     }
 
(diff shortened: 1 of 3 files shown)


altic-dev:
Are you using this branch now? I had some of the delays to make sure it avoid race conditions and crashes while initiating. Wasn't there without any reason most of the time. Would love to know how it works for you overall before we can think about merging this actually. Will affect a lot of users if done wrong haha.

altic-dev:
Closing as stale PR. If you want to work on this, pleas open this again!

PR #191, opened 2026-03-05
fix(ui): smooth bottom overlay animation and fix border clipping

floze-the-genius:
## Summary

- Fixes border/corner clipping on the bottom overlay when it appears
- Replaces janky window-level animation with smooth GPU-accelerated SwiftUI animation
- Forces correct initial layout before showing the overlay window

## Problems fixed

**Border clipping**: The overlay window was sized exactly to the SwiftUI content bounds, causing the 1px border stroke and rounded corners to be clipped at the window edge. Fixed by adding 4pt outer padding.

**Janky appear/disappear animation**: The previous implementation used `NSAnimationContext` to animate the window's position (`setFrameOrigin`), which runs at the window-server level and produces visible stuttering. Replaced with SwiftUI-internal spring animation (`scaleEffect` + `opacity` + `offset`) that runs GPU-accelerated within the compositor.

**Incorrect initial size**: On first show, the window could appear with wrong dimensions because SwiftUI hadn't completed its layout pass. Fixed by calling `layoutSubtreeIfNeeded()` and re-measuring `fittingSize` before making the window visible.

## Changes

**`BottomOverlayView.swift`**:
- `show()`: Force layout pass, set correct content size, show window with alpha=1 immediately (no NSAnimationContext), signal SwiftUI via `bottomOverlayPresented`
- `hide()`: Signal SwiftUI to animate out, then hide window after 200ms delay
- Body: Added `@State isPresented` with spring animation (scale 0.92→1, opacity 0→1, offset y 8→0, response 0.28s, damping 0.82)
- Body: Added `.padding(4)` for breathing room

**`NotchContentViews.swift`**:
- Added `@Published var bottomOverlayPresented: Bool` to `NotchContentState` for coordinated appear/disappear

## Animation parameters

```swift
.scaleEffect(isPresented ? 1 : 0.92)
.opacity(isPresented ? 1 : 0)
.offset(y: isPresented ? 0 : 8)
.animation(.spring(response: 0.28, dampingFraction: 0.82), value: isPresented)
```

### DIFF
--- a/Sources/Fluid/Views/BottomOverlayView.swift
+++ b/Sources/Fluid/Views/BottomOverlayView.swift
@@ -78,18 +78,23 @@ final class BottomOverlayWindowController {
             self.createWindow()
         }
 
+        guard let window = self.window, let hostingView = window.contentView as? NSHostingView<BottomOverlayView> else { return }
+
+        // Force a layout pass so fittingSize is accurate before we show
+        hostingView.layoutSubtreeIfNeeded()
+        let fittingSize = hostingView.fittingSize
+        hostingView.frame = NSRect(origin: .zero, size: fittingSize)
+        window.setContentSize(fittingSize)
+
         // Position at bottom center of main screen
         self.positionWindow()
 
-        // Show with animation
-        self.window?.alphaValue = 0
-        self.window?.orderFrontRegardless()
+        // Show window instantly — SwiftUI handles the visual appear animation
+        window.alphaValue = 1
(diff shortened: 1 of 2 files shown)


grohith327:
Hey @guard22 Can you share a recording of the new animation ?

altic-dev:
Hi! Thanks for the video but what does it show specifically? is it after the fix or before the fix? Confused - sorry.

PR #193, opened 2026-03-07
feat: support Globe/fn key as hotkey without triggering Emoji Picker

Yves000:
## Summary

- Move CGEvent tap to a dedicated high-priority thread so events are intercepted before macOS triggers system actions (like the Emoji Picker)
- Add `GlobeKeyManager` service that suppresses/restores the system Globe key action (`AppleFnUsageType`) via `CFPreferences` + `DistributedNotification`
- Consume synthetic keyCode 179 events (the actual trigger for the Emoji Picker)
- Crash recovery: if the app is force-quit while the override is active, the original Globe key setting is restored on next launch
- Show info text in Settings when a Globe key shortcut is active

## Context

Using the Globe/fn key (keyCode 63) as a dictation hotkey caused the macOS Emoji Picker to open alongside FluidVoice's dictation every time the key was pressed. Dictation itself worked fine (start/stop), but the Emoji Picker appearing in parallel was disruptive.

## Test plan

- [ ] Set Globe key as Transcribe Mode shortcut → Emoji Picker should NOT appear
- [ ] Globe key press → dictation starts
- [ ] Globe key press again → dictation stops
- [ ] Press and Hold mode: hold Globe → dictation starts, release → stops
- [ ] Change shortcut to something else → Emoji Picker works again
- [ ] Force-quit app while Globe key shortcut is active → on next launch, Globe key system setting is restored
- [ ] Other hotkeys (Ctrl, Cmd, etc.) continue to work correctly

### DIFF
--- a/Sources/Fluid/AppDelegate.swift
+++ b/Sources/Fluid/AppDelegate.swift
@@ -55,6 +55,8 @@ class AppDelegate: NSObject, NSApplicationDelegate {
 
     func applicationWillTerminate(_ notification: Notification) {
         DebugLogger.shared.info("Application will terminate", source: "AppDelegate")
+        // Restore Globe key system setting before quitting
+        GlobeKeyManager.shared.restoreSystemGlobeAction()
         // Clean up the update check timer
         self.updateCheckTimer?.invalidate()
         self.updateCheckTimer = nil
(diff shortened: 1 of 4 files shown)


altic-dev:
If you want to try to merge this, would appreciate if you follow the PR template + show some images,video of the fix and checking out codex PR suggestions :) Thanks

altic-dev:
image


figured this part is in keyboard settings you can disable it there natively. Closing this as we dont want FV to control this for users. If it has a different intention, please feel free to open it! Thanks

Yves000:
Sorry, I've been really busy these past few weeks. I'm sorry for the late reply. The intent of this PR is different from what you think. I wanted to replicate the same behavior as in WhisprFlow, where you can simply use the Globe key as the dictation key without changing anything in macOS settings, and when you close the app, it continues to work normally. But if you think it’s a better solution for users to manually change the behavior via macOS System Preferences, then the PR can remain closed.

PR #229, opened 2026-03-26
perf: reduce perceived recording start delay by ~50%

JereSalo:
## Summary

Reduces the perceived delay between pressing the hotkey and seeing the recording overlay by ~50%, from ~850ms to ~400-450ms. Two independent optimizations:

**1. Show overlay in parallel with engine startup** (`1c2ca54`)

Previously, the overlay expand animation only started after the audio engine was fully initialized, because it was triggered by `isRunning` becoming `true`. This adds `showOverlayEarly()` which fires the overlay immediately on hotkey press, so the ~300ms expand animation runs concurrently with the ~330ms engine startup instead of sequentially after it. The overlay doesn't need the engine to be ready — it only needs the `audioLevelPublisher`, which simply emits no values until the engine starts capturing. Applied to dictation, command, and rewrite mode callbacks.

**2. Keep audio engine alive between recordings** (`dd2c73d`)

Previously, `stop()` destroyed the `AVAudioEngine` instance and the next `start()` had to recreate it from scratch, including the expensive `inputNode` (~166ms) and `outputNode` (~36ms) instantiation where CoreAudio creates the aggregate device. Now `stop()` just calls `engine.stop()` and keeps the instance in memory. On the next `start()`, the nodes are already instantiated so `configureSession()` is near-instant, and only `prepare()` + `start()` need to run. Device changes between recordings are handled normally because in sync mode the engine follows the system default device, and `prepare()` re-negotiates the audio format with whatever device is currently active.

## Measured results

Engine startup time (hotkey callback to `isRunning = true`):

| Scenario | Time |
|---|---|
| Original v1.5.9 | ~360ms |
| First recording after launch | ~130-190ms |
| Subsequent recordings (engine warm) | **85-140ms** |

Total perceived delay (Option key press to overlay fully visible):

| Scenario | Time |
|---|---|
| Original v1.5.9 | ~850ms |
| With both optimizations | **~400-450ms** |

### DIFF
--- a/Sources/Fluid/ContentView.swift
+++ b/Sources/Fluid/ContentView.swift
@@ -2475,6 +2475,11 @@ struct ContentView: View {
                 self.menuBarManager.setOverlayMode(.dictation)
 
                 guard !self.asr.isRunning else { return }
+
+                // Show overlay immediately so its expand animation runs in parallel
+                // with the engine startup, instead of waiting for isRunning to flip.
+                self.menuBarManager.showOverlayEarly(asrService: self.asr)
+
                 if SettingsStore.shared.enableTranscriptionSounds {
                     TranscriptionSoundPlayer.shared.playStartSound()
                 }
@@ -2498,6 +2503,8 @@ struct ContentView: View {
                 self.menuBarManager.setOverlayMode(.command)
 
                 guard !self.asr.isRunning else { return }
+
+                self.menuBarManager.showOverlayEarly(asrService: self.asr)
 
                 // Start recording immediately for the command
                 DebugLogger.shared.info(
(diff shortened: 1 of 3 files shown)


JereSalo:
Hi! I made this PR just because I upgraded to 1.5.9 and though it's fast once you finish speaking at the beginning I found it kinda sluggish, taking almost a whole second from when I pressed option and I saw the animation to start talking, so I made a few changes regarding this, the most important one is initiating the UI and the engine concurrently so that we don't need to wait for the engine to load till we load the UI. This works fine because the UI initialization is not very fast.
I'll be using this version.

Issue #213, opened 2026-03-19
[🐞 BUG]  pastes from clipboard instead of pasting dictation


emmanagoro:
**Describe the bug**
After I updated Fluid started pasting was copied in my clipboard instead of just outputting the dictation, This may be due to this new reliable paste setting inside of Fluid because it seems after turning it to experimental direct type it doesn't happen anymore although I think This text insertion mode isn't supposed to work like this

**To Reproduce**
1. make sure "reliable paste" is on under settings
2. use your shortcut to dictate while enabled

**Expected behavior**
to operate how it did before where it would just paste in whatever was voice typed into the floating Space.

**Environment:**
 - macOS Version: Sequoia 15.1
 - App Version: 1.5.9
 - Architecture: apple silicon
 - 
**Additional context**

**Crash Logs**
If the app crashed, please attach relevant crash logs

altic-dev:
is it consistently doing it and NEVER paste what you dictate?

altic-dev:
Looking into this. I think it's timing related issue. It doesn't happen to me interestingly. Will try to fix it in 1.5.10!

emmanagoro:
> [#224](https://github.com/altic-dev/FluidVoice/issues/224) - tracking it here! - Fix is already in. will update soon!

i think this solved it!

ill use it a bunch today and send updates thank you!

PR #231, opened 2026-03-27
Improve reliable paste verification and avoid FluidVoice self-target duplication

altic-dev:
## Summary
- improve reliable paste verification by using AppleScript-based readback for scriptable apps like Xcode and Notes, while keeping Accessibility verification as the general fallback
- keep clipboard restore safer by waiting for stronger verification before restoring the previous pasteboard contents
- avoid double insertion when FluidVoice itself is frontmost by skipping clipboard copy and external typing in the onboarding/playground self-target case
- bump the app version to 1.5.10

## Validation
- `swiftformat --config .swiftformat Sources`
- `swiftlint --strict --config .swiftlint.yml Sources/`
- `./build_incremental.sh`

## Notes
- local release notes for 1.5.10 were updated separately and intentionally not included in this PR

### DIFF
--- a/Sources/Fluid/ContentView.swift
+++ b/Sources/Fluid/ContentView.swift
@@ -1233,6 +1233,22 @@ struct ContentView: View {
         )
     }
 
+    private func resolveTypingTargetPID() -> (pid: pid_t?, shouldRestoreOriginalFocus: Bool) {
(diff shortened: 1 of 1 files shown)


altic-dev:
#213  covered!

PR #278, opened 2026-04-14
Pad short audio with trailing silence instead of dropping it

### LINKED ISSUE #276: [🐞 BUG] Short utterances (<1s) are silently dropped by transcription guard
**Describe the bug**

Any recording shorter than 1 second of audio is silently discarded before transcription even runs. Short single-word utterances like "yes", "no", "stop", "undo", "next" never make it into the transcript and appear to the user as if nothing happened — there's no error, no toast, no history entry, just silence.

The guard lives in `Sources/Fluid/Services/ASRService.swift` around line 933:

```swift
// Avoid whisper.cpp assertions by skipping too-short audio buffers
let minSamples = 16_000
guard pcm.count >= minSamples else {
    DebugLogger.shared.debug(
        "stop(): insufficient audio for transcription (\(pcm.count)/\(minSamples) samples)",
        source: "ASRService"
    )
    ...
    return ""
}
```

The comment scopes the workaround to whisper.cpp, but the guard runs unconditionally inside `stop()` — so even users on Parakeet Flash / v3 / v2, Cohere Transcribe, or Apple Speech (none of which have this assertion issue) lose their sub-1s utterances too.

**To Reproduce**

1. Set the speech model to anything non-Whisper (e.g. Parakeet TDT v3 or Apple Speech).
2. Press the hotkey and immediately say a short word ("yes") for well under a second.
3. Release the hotkey.
4. Observe: nothing is typed, nothing appears in history, no error is shown.

**Expected behavior**

Short utterances should transcribe normally on providers that can handle them. At minimum:
- Only apply the 1s guard when the active provider actually requires it (whisper.cpp).
- For providers that do require a minimum, give the user *some* feedback rather than failing silently — a toast, a subtle overlay flash, or an entry in history flagged as "too short". Silent drops are the worst-case UX because the user can't tell whether the hotkey, the mic, or the transcription failed.

Ideally the minimum would also be user-configurable (or at least lower — 300-500 ms covers most single-word commands without tripping whisper.cpp assertions on buffers that are genuinely empty).

This matters a lot for Command Mode, where short affirmations / corrections / single-word commands are a common and natural input pattern.

**Environment:**
- macOS Version: 26.3.1 (arm64)
- App Version: 1.5.12
- Architecture: Apple Silicon

**Additional context**

Found while using Command Mode — short confirmations kept being ignored with no feedback. Tracked it down to this guard.

domdomegg:
## Summary

Fixes #276.

Sub-1s recordings currently return an empty string silently — short single-word utterances like \"yes\", \"no\", \"stop\" never reach the transcription provider and the user sees no feedback that anything went wrong. The guard was added to dodge a whisper.cpp assertion on buffers shorter than 1s, but it applied to every provider, so Parakeet / Apple Speech / Cohere lost short utterances too.

This PR replaces the early-return with zero-padding:

- Empty buffer (no audio captured at all) → still an early-return, since padding zero samples is pointless.
- Non-empty buffer shorter than 16,000 samples (1s @ 16 kHz) → pad with trailing silence up to 16,000 samples before handing to the provider.

Padding unconditionally (rather than branching per provider) keeps the code simple — whisper.cpp gets the 1s minimum it needs, and every other provider just sees a brief silent tail after the speech, which they all handle fine.

No API surface changes; the only observable behaviour change is that short utterances now transcribe instead of being dropped.

## Test plan

- [ ] Set the speech model to Parakeet TDT v3 (or Apple Speech). Press-and-release the hotkey quickly while saying a single short word (\"yes\"). Confirm the transcription appears rather than nothing.
- [ ] Set the speech model to Whisper Base. Repeat the same short-word test — confirm it still works and no whisper.cpp assertion fires.
- [ ] Long (>1s) recordings still transcribe normally across all providers (no regression).
- [ ] Press-and-release the hotkey so fast that no audio is captured — confirm the debug log shows \"no audio captured, skipping transcription\" and nothing is typed (same as before).

## Notes

- I couldn't run `xcodebuild` locally (Command Line Tools only, no full Xcode). The diff is small and mechanical — please verify CI is green.

### DIFF
--- a/Sources/Fluid/Services/ASRService.swift
+++ b/Sources/Fluid/Services/ASRService.swift
@@ -927,22 +927,35 @@ final class ASRService: ObservableObject {
 
         // NOW it's safe to access the buffer - all pending tasks have completed
         // Thread-safe copy of recorded audio
-        let pcm = self.audioBuffer.getAll()
+        var pcm = self.audioBuffer.getAll()
         self.audioBuffer.clear()
 
-        // Avoid whisper.cpp assertions by skipping too-short audio buffers
-        let minSamples = 16_000
-        guard pcm.count >= minSamples else {
+        // Drop recordings with no audio at all — nothing to transcribe.
+        guard !pcm.isEmpty else {
             DebugLogger.shared.debug(
-                "stop(): insufficient audio for transcription (\(pcm.count)/\(minSamples) samples)",
+                "stop(): no audio captured, skipping transcription",
                 source: "ASRService"
             )
-            // Resume media playback if we paused it
             if shouldResumeMedia {
                 await MediaPlaybackService.shared.resumeIfWePaused(true)
-                DebugLogger.shared.info("🎵 Resumed system media after insufficient audio", source: "ASRService")
+                DebugLogger.shared.info("🎵 Resumed system media after empty audio", source: "ASRService")
             }
             return ""
+        }
+
+        // Pad sub-1s buffers with trailing silence so short utterances (e.g.
+        // "yes", "stop") still transcribe. whisper.cpp asserts on buffers
+        // shorter than 1s; every other provider handles silence padding
+        // without issue, so we pad unconditionally rather than branching per
+        // provider.
+        let minSamples = 16_000
+        if pcm.count < minSamples {
+            let originalCount = pcm.count
+            pcm.append(contentsOf: repeatElement(0.0, count: minSamples - pcm.count))
+            DebugLogger.shared.debug(
+                "stop(): padded short audio with silence (\(originalCount) → \(pcm.count) samples)",
+                source: "ASRService"
+            )
         }
 
         do {


domdomegg:
Okay I've tested this one locally and it works well :)

domdomegg:
@altic-dev can you look over this one? Would be a meaningful improvement as I've been finding it cuts out a lot on short phrases, and this fix works well.

altic-dev:
Hey! I already raised it as an upstream issue to FluidAudio but this actually is a very nice hack. @grohith327 wanna take a look at this when possible?

grohith327:
Yep, will get to it tonight

altic-dev:
Brilliant fix - Just tried it out. Even the UI delay issues from earlier is solve by this actually. The claim of 'UI is slow to pop up, so ASR misses first second' is also cleared. I think we'll stick with this hack instead of waiting for the upstream fix to reduce the actual value tbh. Merging it man! Thanks. Little busy with week - so taking a while to make merges.

PR #279, opened 2026-04-14
Throw AI post-processing errors instead of typing them into the user's document

### LINKED ISSUE #242: [🐞 BUG] AI post-processing errors typed into user's document instead of falling back to raw transcription
When AI post-processing is enabled, if the underlying LLM API call fails, the error message is returned as the "final text" and typed directly into the user's active application.

For example, a user dictating into a document could see `Error: HTTP 503: This model is currently experiencing high demand. Spikes in demand are usually temporary. Please try again later.` inserted instead of their spoken words.

**Steps to reproduce:**
1. Enable AI post-processing in settings
2. Trigger any LLM error (model overloaded, network issue, invalid API key, etc.)
3. Start dictation, speak a sentence, and complete dictation
4. Observe the error string typed into the active app instead of the transcribed text

**Affected code paths in `ContentView.processTextWithAI()`:**
- Empty API response → returns `"<no content>"`
- Any LLM error → returns `"Error: {description}"`

**Expected behavior:** When AI processing fails, the app should gracefully degrade and return the raw transcription to the user. The error should be logged for debugging but never surfaced as typed text.

domdomegg:
## Summary

Fixes #242.

When AI post-processing failed (rate limit, network blip, missing API key, empty response, etc.), `processTextWithAI` converted the failure into a String like `"Error: HTTP 503: This model is currently experiencing high demand…"` and the dictation callsites treated that string as the text to type. Users saw the error message inserted into whatever app they were dictating into.

### Changes

- **`ContentView.processTextWithAI` now throws.** Return type becomes `async throws -> String`. The three failure modes throw instead of string-returning: missing API key, empty response, and underlying `LLMClient` errors propagate directly.
- **Added `AIProcessingError`** (a `LocalizedError`) for the two synthesised failure cases so the catch sites get a useful `localizedDescription` to log / display.
- **`AppleIntelligenceProvider.process` also throws now.** It had the same `return "Error: …"` pattern on FoundationModels errors, so the AI path via Apple Intelligence had the same bug. Made it `async throws -> String` too and let its single caller propagate.
- **Four callsites updated:**
  | Callsite | Behaviour on error |
  |---|---|
  | Dictation (main flow) | Fall back to the raw transcription, log the error |
  | Dictation (reprocess flow) | Same — fall back to raw transcription |
  | Prompt-test UI | Surface error in the existing `promptTest.lastError` slot (matches the already-handled 'not configured' case) |
  | AI chat tab (`callOpenAIChat`) | Keep showing `"Error: …"` since that panel is purely display, not typed anywhere |

The bug report suggested gracefully degrading to raw transcription and logging the error — that's what the dictation callsites now do. The display-only callsites keep showing errors because for them the error text is informative, not destructive.

## Test plan

- [ ] Configure a valid provider + model but set an obviously wrong API key (e.g. `"sk-invalid"`). Dictate a sentence into a text field. Confirm the raw transcription is typed instead of `"Error: HTTP 401 …"`, and the log contains `"AI post-processing failed, falling back to raw transcription: …"`.
- [ ] Remove the API key entirely (blank) and dictate. Confirm the raw transcription is typed and the log shows `"API key not set for …"`.
- [ ] Force an empty model response (e.g. via a test provider returning `"choices": []`). Confirm fallback behaviour.
- [ ] Prompt-test UI: trigger an error and confirm the error banner appears in the testing card (no output written to `lastOutputText`).
- [ ] AI chat tab: trigger an error and confirm it's still shown in `aiOutputText` as before (display-only, no regression).
- [ ] Happy path still works end-to-end across OpenAI / Anthropic / Groq / Apple Intelligence.

## Notes

- I couldn't run `xcodebuild` locally (Command Line Tools only, no full Xcode). Please verify CI is green — the diff is mechanical but the throw-propagation touches a few signatures so it's worth a compile pass.
- `OpenAICompatibleProvider` in `AIProvider.swift` has the same string-error pattern but is dead code (declared and never referenced) — left alone for this PR.

### DIFF
--- a/Sources/Fluid/ContentView.swift
+++ b/Sources/Fluid/ContentView.swift
@@ -13,6 +13,22 @@ import CoreGraphics
 import Security
 import SwiftUI
 
+// MARK: - AI Processing Errors
+
+enum AIProcessingError: LocalizedError {
+    case missingAPIKey(provider: String)
+    case emptyResponse
+
+    var errorDescription: String? {
+        switch self {
+        case .missingAPIKey(let provider):
+            return "API key not set for \(provider)"
+        case .emptyResponse:
+            return "AI returned an empty response"
+        }
+    }
+}
+
 // MARK: - Sidebar Item Enum
 
 enum SidebarItem: Hashable {
@@ -1491,7 +1507,7 @@ struct ContentView: View {
         _ inputText: String,
         overrideSystemPrompt: String? = nil,
         dictationSlot: SettingsStore.DictationShortcutSlot? = nil
-    ) async -> String {
+    ) async throws -> String {
         // CRITICAL FIX: Read current settings from SettingsStore, not stale @State copies
         // This ensures AI provider/model changes in AISettingsView take effect immediately
         let currentSelectedProviderID = SettingsStore.shared.selectedProviderID
@@ -1564,7 +1580,7 @@ struct ContentView: View {
                     self.logDictationPromptTrace("Selected context text", value: "<none (dictation mode)>")
                 }
                 DebugLogger.shared.debug("Using Apple Intelligence for transcription cleanup", source: "ContentView")
-                let output = await provider.process(systemPrompt: systemPrompt, userText: inputText)
+                let output = try await provider.process(systemPrompt: systemPrompt, userText: inputText)
                 if self.shouldTracePromptProcessing {
                     self.logDictationPromptTrace("Model answer (A)", value: output)
                 }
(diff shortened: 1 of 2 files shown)


domdomegg:
This works locally, I've tested by disabling an AI key. The raw transcription is pasted and the logs include the 401 error.

domdomegg:
One trade-off is that errors may be less visible though to end users. Gonna see if we can show some kind of error toast in a sensible way.

domdomegg:
Okay I think this is good as-is. Going to do a separate PR (in https://github.com/altic-dev/FluidVoice/pull/281) for recording and surfacing errors better.

grohith327 (approved):
LGTM

altic-dev:
Merged via #303 with current main conflict resolved.

PR #281, opened 2026-04-14
Flag AI post-processing fallbacks in transcription history (builds on #279)

domdomegg:
## Summary

Stacked on top of #279. Please merge that first — this PR's diff will shrink to just the history-flag changes once it does, but GitHub can't target a head branch that isn't merged yet, so for now the diff here shows both sets of changes.

The problem: after #279 a failed AI cleanup silently falls back to the raw transcription, with only a line in `~/Library/Logs/Fluid/Fluid.log`. A broken API key or expired credit can go unnoticed for days.

### Durable history flag

Added `aiProcessingError: String?` to `TranscriptionHistoryEntry` (Codable via `decodeIfPresent` — fully back-compat with existing persisted history, old entries decode with `nil`). Threaded the fallback reason into `addEntry` at the two dictation callsites (main dictation flow and the reprocess flow).

UI:

- **History list row**: orange ⚠ triangle next to entries where `aiProcessingError` is set; hover reveals the error message.
- **History detail view**: orange callout card at the top of the detail pane with the full error message.

Screenshot in the comment below.

### What I dropped

Initially I also tried surfacing failures as a transient message in the notch / bottom overlay ("⚠︎ AI cleanup failed — typed raw", held for ~4s). I couldn't get it to actually persist long enough to be readable — the post-typing `NotchOverlayManager.hide()` fires immediately after typing completes and overrode the transient status each time. Left that out of the final PR; the history flag is the durable signal and is the part that worked reliably. Happy to revisit the overlay approach in a follow-up if maintainers have a cleaner hook point.

## Notes

- Could not run CI locally (CLT was insufficient — needed full Xcode). Built and ran on Apple Silicon / macOS 26.3.1 with Xcode 26.4 to verify. One unrelated local-only change (bumping `modelcontextprotocol/swift-sdk` from 0.10.2 → 0.12.0 in `Package.resolved`) was needed to get past a Swift 6 concurrency error in the MCP transport — not included in this PR since CI on Xcode 26.3 doesn't need it.
- Orange colour for the warning is hardcoded; could be themed if maintainers prefer.

### DIFF
--- a/Sources/Fluid/ContentView.swift
+++ b/Sources/Fluid/ContentView.swift
@@ -1851,6 +1851,7 @@ struct ContentView: View {
         }
 
         var finalText: String
+        var aiFallbackReason: String?
 
         let shouldUseAI = activeDictationSlot.map { DictationAIPostProcessingGate.isConfigured(for: $0) } ??
             DictationAIPostProcessingGate.isConfigured()
@@ -1881,6 +1882,7 @@ struct ContentView: View {
                     "AI post-processing failed, falling back to raw transcription: \(error.localizedDescription)",
                     source: "ContentView"
                 )
+                aiFallbackReason = error.localizedDescription
                 finalText = transcribedText
             }
             let postProcessingLatencyMs = Int((Date().timeIntervalSince(postProcessingStart) * 1000).rounded())
@@ -1946,7 +1948,8 @@ struct ContentView: View {
                 rawText: transcribedText,
                 processedText: finalText,
                 appName: appInfo.name,
-                windowTitle: appInfo.windowTitle
+                windowTitle: appInfo.windowTitle,
+                aiProcessingError: aiFallbackReason
             )
         }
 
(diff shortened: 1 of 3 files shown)


domdomegg:
CleanShot 2026-04-14 at 06 29 31@2x

grohith327 (approved):
LGTM!

grohith327:
@altic-dev @domdomegg An idea for the future, showing errors as a default macos notification or a compact notch notification (based on whether the active screen has a notch or not) would be a good way to let the users know immediately what the error was. Let me know what you guys think!

domdomegg:
> An idea for the future, showing errors as a default macos notification or a compact notch notification (based on whether the active screen has a notch or not) would be a good way to let the users know immediately what the error was. Let me know what you guys think!

Yeah I agree with this! I did briefly look into it but I think it requires setting up notification permissions which made it a bit more of an involved PR. Happy to give it a go after we get these two merged.

PR #286, opened 2026-04-17
Drop temperature for Anthropic models that deprecate it

### LINKED ISSUE #285: [🐞 BUG] Opus 4.7 fails with Anthropic provider — `temperature` is deprecated for this model
**Describe the bug**
Selecting Claude Opus 4.7 (`claude-opus-4-7`) with the Anthropic provider causes every request to fail. FluidVoice sends a `temperature` parameter, but Opus 4.7 no longer accepts it.

**To Reproduce**
1. Configure the Anthropic provider with a valid API key
2. Select the Opus 4.7 model (`claude-opus-4-7`)
3. Trigger any transcription/post-processing request
4. See error:
```
Error: HTTP 400: {"error":{"code":"invalid_request_error","message":"temperature is deprecated for this model.","type":"invalid_request_error","param":null}}
```

**Expected behavior**
Request succeeds. FluidVoice should omit `temperature` (or only send it when the selected model supports it) for models that have deprecated it, such as Opus 4.7.

**Environment:**
 - macOS Version: 26.3.1 (build 25D2128)
 - App Version: 1.5.12 (build 10)
 - Architecture: Apple Silicon (arm64)

**Additional context**
Anthropic has deprecated `temperature` on newer models (Opus 4.7 and other extended-thinking-by-default models). Fix is to drop the parameter from the Anthropic request payload (at least for affected models).

domdomegg:
## Summary
- Claude Opus 4.7 rejects the `temperature` parameter (it uses extended thinking by default), so every request sent to it via the Anthropic provider fails with `HTTP 400: temperature is deprecated for this model`.
- Adds `SettingsStore.isTemperatureUnsupported(_:)` covering reasoning models (o1/o3/gpt-5/...) plus Claude Opus 4.7+, and uses it at the three live call sites that build an `LLMClient.Config` (`ContentView`, `CommandModeService`, `RewriteModeService`).
- `isReasoningModel` still gates `max_completion_tokens` and the reasoning token budget — those are OpenAI-specific and shouldn't fire for Anthropic models.

Fixes #285.

## Test plan
- [ ] Select the Anthropic provider with model `claude-opus-4-7` and confirm transcription cleanup, Command Mode, and Rewrite Mode no longer return the HTTP 400 temperature error.
- [ ] Regression-check an OpenAI reasoning model (e.g. `gpt-5`) still sends no `temperature`.
- [ ] Regression-check a non-reasoning model (e.g. `gpt-4.1`, `claude-sonnet-4-20250514`) still sends `temperature: 0.2 / 0.1 / 0.7`.

### DIFF
--- a/Sources/Fluid/ContentView.swift
+++ b/Sources/Fluid/ContentView.swift
@@ -1615,9 +1615,8 @@ struct ContentView: View {
             self.logDictationPromptTrace("Selected context text", value: "<none (dictation mode)>")
         }
 
-        // Check if this is a reasoning model that doesn't support temperature parameter
-        let modelLower = derivedSelectedModel.lowercased()
-        let isReasoningModel = modelLower.hasPrefix("o1") || modelLower.hasPrefix("o3") || modelLower.hasPrefix("gpt-5")
+        // Check if this model doesn't support the temperature parameter
+        let isTemperatureUnsupported = SettingsStore.shared.isTemperatureUnsupported(derivedSelectedModel)
 
         // Get reasoning config for this model (uses per-model settings or auto-detection)
         // This handles custom parameters like reasoning_effort, enable_thinking, etc.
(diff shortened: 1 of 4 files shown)


grohith327 (approved):
LGTM

PR #290, opened 2026-04-19
Add MCP support to command mode

### LINKED ISSUE #275: [✨ FEATURE] MCP server support in Command Mode
**Is your feature request related to a problem?**

Command Mode is great for the built-in Mac actions it already knows how to take, but the set of things it can do is fixed. A lot of the workflows I'd most like to drive by voice live outside of what FluidVoice itself can see — my browser, my inbox, my smart home, my local filesystem, etc. Today the only way to reach those is to fall back to the system (switch apps, type, click) which defeats a lot of the appeal of hands-free voice control.

**Describe the solution you'd like**

Add [Model Context Protocol](https://modelcontextprotocol.io/) client support to Command Mode, so users can attach any MCP server (stdio or HTTP/SSE) and have its tools surfaced to the model alongside the built-in command tools. In practice this would look like:

- A Settings pane where users add MCP servers (command + args for stdio, or URL + auth header for HTTP).
- A per-server on/off toggle and a tool-level allow list, so a noisy server doesn't flood the model's tool list or burn context.
- Tool calls rendered in the existing tool-call UI in Command Mode (same `ToolCall` / `toolOutputView` path already used for built-in tools).
- The existing "Apple Intelligence disabled for Command Mode (no tool support)" guard already tells me the plumbing for tool-using providers is in place — this would extend that surface.

**Concrete workflows this unlocks**

These are all things I'd genuinely use if it shipped:

1. **Browser control (Chrome DevTools MCP / Playwright MCP).** "Open the three tabs I had pinned yesterday and log in to the second one." "Find the checkout button on this page and click it." "Summarise what's on screen and email it to me." Voice is a much better input modality than a trackpad for this kind of multi-step web task.
2. **Email triage (Gmail MCP).** "What emails came in from my bank this week?" "Draft a reply to the last message from Alice saying I'll get back to her on Monday." "Archive everything from newsletters@ that's older than 30 days." Being able to triage the inbox while making coffee is the dream.
3. **Calendar + contacts.** "Do I have anything free tomorrow afternoon?" "Schedule a 30-min call with Bob next Tuesday." "What's Bob's phone number?" These are small, frequent questions where opening Calendar.app is friction.
4. **Smart home (Home Assistant MCP).** "Turn the office lights off and start the robot vacuum." "Is the front door locked?" Obvious voice use case, and unlike Siri it can chain multiple actions in one utterance.
5. **Local shell / filesystem.** "Find the most recent screenshot on my desktop and open it." "Kill whatever's holding port 3000." "Show me the last 10 commits on this branch." Voice is surprisingly nice for short, well-scoped shell-ish tasks.
6. **Project-specific tools.** Anyone who already runs a personal MCP server (notes, task tracker, Linear, Jira, internal APIs…) gets those same tools available by voice for free.

The unifying theme: Command Mode today is a voice UI for FluidVoice's tools; with MCP it becomes a voice UI for *anything the user has plugged in*. That's a step-change in what the feature is useful for.

**Context efficiency: tool search and sandboxed execution**

One thing worth designing in from day one: MCP servers can expose hundreds of tools each, and loading every schema into every turn is both slow and expensive (most of the context window ends up as tool JSON the model never uses). Two patterns that Claude Code already uses well and that would fit Command Mode naturally:

- **Tool search / lazy loading.** Instead of dumping all MCP tool schemas up front, expose a single `tool_search` meta-tool that lets the model query for tools by name or keyword and only load the schemas it actually needs. For a user with Gmail + Calendar + Drive + Home Assistant + Chrome MCP servers attached, this is the difference between ~300 tools in context per turn and ~5. Huge win for latency, cost, and accuracy (smaller tool lists → better tool selection).
- **Sandboxed code execution as an MCP bridge.** A single `execute_code` tool that can invoke other MCP tools inside a sandboxed JS/Python runtime is enormously more efficient than round-tripping every intermediate value through the model. Example: "take a photo from my desktop, run it through the barcode scanner, save the decoded URL to a note" — with a sandbox that's one tool call; without it, the raw image bytes pass through context as base64 (slow, expensive, often blows the window). It also lets the model loop/filter/map over tool results without each iteration hitting the model.

Both techniques are additive to the basic MCP client — they can land in a v2 — but if the initial design assumes "every tool in context every turn" it'll be hard to retrofit.

**Describe alternatives you've considered**

- **Hard-coding each integration.** Doesn't scale — every new integration is a PR against FluidVoice, and users can't add private/internal tools.
- **Piping commands out to Claude Code / a separate agent.** Works, but adds latency, a second UI, and a second place to configure tools. Keeping it in Command Mode is a much tighter UX.
- **AppleScript / Shortcuts bridges.** Covers some of these cases on macOS specifically, but it's Mac-only, brittle, and doesn't match the ecosystem momentum MCP now has (Claude, Cursor, Zed, VS Code, Raycast, etc. all speak it).

**Additional context**

- Anthropic's SDK (which FluidVoice already uses for Claude-backed enhancement) has first-class MCP support, so the lift on the model side is small — the main work is config UI + process lifecycle management for stdio servers.

grohith327:
## Description
This PR adds a bunch a features in order to support MCP for command mode execution. The list of the features are:

- Add separate strategies for routing Anthropic requests and OpenAI compatible requests. 
  - If the model provider is anthropic, messages api is used. Else responses api is preferred and completions is used as fallback
- Minor updates to the command mode chat UI - moved around the buttons a little bit
- Added a mechanism to define the MCP config and edit it from the FluidVoice app
- Fixed some issues on tool calls are rendered on the chat
- Once action is completed mode, depending on whether the user active display screen has notch - a compact dynamic island notification is shown for success, else a system notification is sent for success
- Added Markdown syntax rendering in the command mode chat

## Type of Change
- [ ] 🐞 Bug fix (non-breaking change which fixes an issue)
- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 📝 Documentation update

## Related Issues
- Closes #275 

## Testing
- [x] Tested on macOS [version]
- [x] Built locally: `sh build_incremental.sh`

## Notes
- Due to the size of the change, would recommend pulling the code and testing once. 
- Do a beta release first

### DIFF
--- a/Fluid.xcodeproj/project.pbxproj
+++ b/Fluid.xcodeproj/project.pbxproj
@@ -17,6 +17,7 @@
 		7CDB0A2F2F3C4D5600FB7CAD /* dictation_fixture.wav in Resources */ = {isa = PBXBuildFile; fileRef = 7CDB0A2B2F3C4D5600FB7CAD /* dictation_fixture.wav */; };
 		7CDB0A302F3C4D5600FB7CAD /* XCTest.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 7CDB0A2C2F3C4D5600FB7CAD /* XCTest.framework */; };
 		7CE006BD2E80EBE600DDCCD6 /* AppUpdater in Frameworks */ = {isa = PBXBuildFile; productRef = 7CE006BC2E80EBE600DDCCD6 /* AppUpdater */; };
+		7CF2A0012F7C100100A1B2C3 /* MarkdownUI in Frameworks */ = {isa = PBXBuildFile; productRef = 7CF2A0032F7C100100A1B2C3 /* MarkdownUI */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
@@ -51,6 +52,7 @@
 			isa = PBXFrameworksBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
+				7CF2A0012F7C100100A1B2C3 /* MarkdownUI in Frameworks */,
 				7C3697892ED70F9C005874CE /* DynamicNotchKit in Frameworks */,
 				7C5AF14B2F15041600DE21B0 /* MediaRemoteAdapter in Frameworks */,
 				7C3236542EAAD608007E4CB6 /* MCP in Frameworks */,
@@ -157,6 +159,7 @@
 				7CE006BC2E80EBE600DDCCD6 /* AppUpdater */,
 				7C1C72F12EECBD1300E3BF4D /* SwiftWhisper */,
 				7C5AF14A2F15041600DE21B0 /* MediaRemoteAdapter */,
+				7CF2A0032F7C100100A1B2C3 /* MarkdownUI */,
 			);
 			productName = FluidVoice;
 			productReference = 7C078D8F2E3B339200FB7CAC /* FluidVoice Debug.app */;
@@ -214,6 +217,7 @@
 				7C3697872ED70F9C005874CE /* XCRemoteSwiftPackageReference "DynamicNotchKit" */,
 				7C1C72F02EECBD1300E3BF4D /* XCRemoteSwiftPackageReference "SwiftWhisper" */,
 				7C5AF1492F15041600DE21B0 /* XCRemoteSwiftPackageReference "mediaremote-adapter" */,
+				7CF2A0022F7C100100A1B2C3 /* XCRemoteSwiftPackageReference "swift-markdown-ui" */,
 			);
 			preferredProjectObjectVersion = 77;
 			productRefGroup = 7C078D902E3B339200FB7CAC /* Products */;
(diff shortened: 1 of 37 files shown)


altic-dev:
One blocker before merge: chat completions fallback still receives Responses-style tool definitions.

In `buildChatCompletionsBody`, `body["tools"] = config.tools` forwards tools shaped like `{"type":"function","name":...,"parameters":...}`. If `/v1/responses` is rejected and we fall back to `/v1/chat/completions`, Chat Completions expects `{"type":"function","function":{"name":...,"parameters":...}}`, so command mode tool calls can 400 or be ignored exactly on fallback providers.

Fix should be small: add a `normalizeToolsForChatCompletions(...)` helper and use it in `buildChatCompletionsBody`, while keeping `normalizeToolsForResponses(...)` for Responses.

altic-dev:
One more concrete blocker I found in MCP tool registration: `uniqueToolName` can loop forever when two tool names collide after the 64-character truncation.

`base` is already sanitized/truncated to 64 chars. On collision, `candidate = sanitizeToolName("\(base)_\(counter)")` appends a suffix, but `sanitizeToolName` truncates back to 64 chars, so the suffix is dropped and `candidate` remains identical to `base` forever.

Scenario: two MCP tools with names that differ only after the first 64 sanitized characters, or long server/tool combinations with the same prefix. That would hang MCP bootstrap/tool catalog rebuild.

Fix should be small: when adding a collision suffix, reserve suffix length before truncating the base, e.g. truncate to `64 - suffix.count` then append `_2`, `_3`, etc.

PR #293, opened 2026-04-21
Stop invalidating ContentView on streaming partial transcription updates

NathanDrake2406:
## What this changes

`ASRService.partialTranscription` moves from `@Published` to a private `CurrentValueSubject<String, Never>` exposed to subscribers via an `AnyPublisher`. The value getter and the single Combine consumer (`MenuBarManager`) are updated accordingly. No other `@Published` properties on `ASRService` are affected.

## Why

During streaming ASR, `partialTranscription` is written 1-5 times per second depending on the model (`parakeetRealtime` runs at 0.2s intervals, default models at 0.6s, Cohere at 1.0s per `SettingsStore.swift`).

Because it was `@Published`, every write fired Combine's synthesized `objectWillChange.send()`. `AppServices.setupASRForwarding` pipes `ASRService.objectWillChange` into `AppServices.objectWillChange`, and `ContentView` observes `AppServices` via `@EnvironmentObject`. SwiftUI's legacy `ObservableObject` protocol has no per-property dependency tracking, so every tick invalidated `ContentView` and forced its 3,061-line body to re-evaluate, even though `ContentView` never reads `partialTranscription`.

`MenuBarManager` is the only consumer of this value (it forwards to the notch overlay when streaming preview is enabled). Moving the hot path off `objectWillChange` eliminates the invalidation storm without touching any legitimate cold-signal forwarding (`isRunning`, `isAsrReady`, `micStatus`, `errorTitle`, etc. stay `@Published` and `ContentView` continues to react to them correctly).

## Approach

1. Replace `@Published var partialTranscription: String = ""` with `private let partialTranscriptionSubject = CurrentValueSubject<String, Never>("")`.
2. Expose `var partialTranscription: String { subject.value }` for synchronous reads and `var partialTranscriptionPublisher: AnyPublisher<String, Never>` for subscription.
3. Update the three writers in `ASRService` to call `partialTranscriptionSubject.send(...)` instead of assigning the property.
4. Update `MenuBarManager.configure(asrService:)` to subscribe via `asrService.partialTranscriptionPublisher` instead of `asrService.$partialTranscription`.

Non-goals: no changes to any other `@Published` property, no migration to `@Observable`, no changes to `AppServices` forwarding or `ContentView`. This is the minimal surgical fix to the documented hot path.

## Validation

- SwiftLint strict: 0 violations on both edited files.
- `xcodebuild build`: clean with respect to this change. (The swift-sdk MCP transport has two pre-existing data-race errors on `upstream/main` that remain; confirmed identical on the unmodified branch.)
- Manual verification of the value flow: `partialTranscriptionSubject.send(x)` updates the getter and the published publisher on the next subscription tick. `MenuBarManager`'s `.receive(on: DispatchQueue.main).sink { ... }` still delivers on main as before.

## Risks / follow-ups

- Any future code that subscribes to `$partialTranscription` via the Combine projected value will break and must use `partialTranscriptionPublisher` instead. Currently there are no other consumers.
- Broader cleanup opportunities remain in the AppServices forwarding design (blanket `objectWillChange` forwarding for `AudioHardwareObserver` still exists, and migrating `ASRService` to the `@Observable` macro would give per-property tracking across all fields). Those are out of scope for this PR.

### DIFF
--- a/Sources/Fluid/Services/ASRService.swift
+++ b/Sources/Fluid/Services/ASRService.swift
@@ -100,7 +100,14 @@ private actor ModelDownloadRegistry {
 final class ASRService: ObservableObject {
     @Published var isRunning: Bool = false
     @Published var finalText: String = ""
-    @Published var partialTranscription: String = ""
+    // Not @Published: writes fire 1-5x/s during streaming ASR. Routing them through
+    // objectWillChange invalidates every ContentView subscriber of AppServices, even
+    // though ContentView never reads this value. MenuBarManager subscribes directly.
+    private let partialTranscriptionSubject = CurrentValueSubject<String, Never>("")
+    var partialTranscription: String { self.partialTranscriptionSubject.value }
+    var partialTranscriptionPublisher: AnyPublisher<String, Never> {
+        self.partialTranscriptionSubject.eraseToAnyPublisher()
+    }
     @Published var wordBoostStatusText: String = "Word boost: off"
     @Published var micStatus: AVAuthorizationStatus = .notDetermined
     @Published var isAsrReady: Bool = false
@@ -749,7 +756,7 @@ final class ASRService: ObservableObject {
         DebugLogger.shared.debug("🧹 Clearing buffers and state", source: "ASRService")
         self.finalText.removeAll()
         self.audioBuffer.clear(keepingCapacity: true) // specific optimization for restart
-        self.partialTranscription.removeAll()
+        self.partialTranscriptionSubject.send("")
         self.previousFullTranscription.removeAll()
         self.lastBoostHitTerm = nil
         self.lastProcessedSampleCount = 0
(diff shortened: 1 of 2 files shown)


altic-dev:
PR is 12 lines and PR summary is a 100+ lines ahahah. can you explain a little on what this is and why did you have to fix it and was there any issue that this curates the fix toward ? Seems like a lot of Claude Slop that is confusing me xD

NathanDrake2406:
> PR is 12 lines and PR summary is a 100+ lines ahahah. can you explain a little on what this is and why did you have to fix it and was there any issue that this curates the fix toward ? Seems like a lot of Claude Slop that is confusing me xD

It's a small perf improvement. 
The motivation for this PR is that the app, while functionally good, is quite laggy compared to sth like superwhisper (i mean ofc, this is free). 

It basically is doing this: avoid unnecessary UI invalidation during streaming ASR. It's not a fix fix, but a small incremental cleanup

PR #325, opened 2026-05-06
Music duck and keep-awake behaviour during dictation

AndrewBeniston:
## What this adds

Two independent settings, both wired through the existing pause-media hooks in `ASRService`. Either could be merged on its own; happy to split into two PRs if you prefer.

### Music During Transcription (replaces the existing pause toggle)

The current `Pause Media During Transcription` boolean becomes a three-mode picker: **Leave Playing**, **Pause** (the existing behaviour, unchanged), or **Lower Volume**.

Lower Volume snapshots the system output volume, fades it down to 10% over 200ms, and fades it back up the same way when you stop. The reason for the duck rather than the existing pause: psychologically it feels much nicer to have the music keep playing quietly. When the audio cuts entirely, the absence of background sound is jarring — a bit like in a film when the room tone drops out and you suddenly notice the silence. Dipping the volume to 10% keeps the ambience without competing with what you're saying into the mic. 10% turns out to be the sweet spot — quiet enough not to corrupt the transcript, loud enough that the room still feels alive.

### Keep Mac Awake While Dictating

Holds an `IOPMAssertion` of type `kIOPMAssertionTypePreventUserIdleDisplaySleep` for the duration of a recording so the screen doesn't dim or lock mid-session. Released the instant recording stops. Useful for long-form dictation where the user is sitting still and the display would otherwise time out and trigger the screen-lock cascade.

Defaults to `true`. Doesn't cover lid-close sleep (that would need `kIOPMAssertionTypePreventSystemSleep` with the lid-close option, which feels like overreach — happy to add it if you want).

## Implementation notes

- `SystemVolumeController` (new) — thread-safe wrapper around `AudioObjectGet/SetPropertyData` on the default output device's master volume scalar. Falls back to writing channels 1+2 individually for devices that don't expose `kAudioObjectPropertyElementMain`.
- `SleepPreventionService` (new) — `IOPMAssertionCreateWithName` / `IOPMAssertionRelease` wrapper.
- `MediaPlaybackService` gains `duckSystemVolume()` / `restoreSystemVolume(previous:)` and a unified `restore(from: MediaSessionAction)` that handles either pause-resume or volume-restore at every exit path.
- The fade ramp runs on a detached `Task` so the main actor isn't blocked between steps. A new fade cancels any in-flight one, which matters when the user releases the hotkey before the duck-down has finished and the fade-up needs to take over cleanly. If the user's volume is already at or below 10%, the duck is skipped entirely (would otherwise raise a quieter volume up to 10%, which is the wrong direction).
- Volume restore fires the moment the audio engine is fully torn down, not at the end of `stop()` after final transcription completes. That way the volume snaps back when the user releases the hotkey rather than 1-3s later when the model finishes.

## Backwards compatibility

The legacy `pauseMediaDuringTranscription` UserDefaults key migrates to a new `MediaBehaviorDuringTranscription` enum on first read. The `pauseMediaDuringTranscription` Bool getter/setter remains on `SettingsStore` as a shim (true ↔ `.pause`, false ↔ `.none`) so the `BackupService` payload stays format-compatible without a schema bump. Old backups restore cleanly as `.pause` or `.none`. The duck setting itself doesn't survive a backup/restore round trip — felt acceptable since it's a UX preference rather than load-bearing config, but happy to bump the backup schema if you'd rather.

## Testing

Built and tested locally on Apple Silicon (macOS 15) against `upstream/main`. Verified manually:

- Picker switches cleanly between modes; legacy boolean migrates on first launch
- Lower Volume + Spotify: smooth fade-down on hotkey hold, smooth fade-up on release; volume returns to original level (not just 100%)
- Lower Volume with system volume already at 5%: skipped, no false volume raise
- Tap-and-immediately-release hotkey: fade-up starts from wherever the fade-down had got to, no jump
- Keep-awake assertion holds while recording, releases on stop (verified via `pmset -g assertions | grep FluidVoice`)
- Keep-awake released even on `start()` failure path (forced by denying mic mid-test)
- Plist round-trip: settings persist across app restarts under the same `com.FluidApp.app` bundle ID

Not added: integration tests for either feature. Felt awkward to test system-volume mutation and `IOPMAssertion` lifecycle in CI, but if you'd like me to attempt it I'll happily add them.

### DIFF
--- a/Sources/Fluid/ContentView.swift
+++ b/Sources/Fluid/ContentView.swift
@@ -2947,6 +2947,25 @@ extension ContentView {
 
     private func beginDictationRecording(for slot: SettingsStore.DictationShortcutSlot, mode: ActiveRecordingMode) {
         DebugLogger.shared.debug("Begin dictation recording for slot \(slot.rawValue)", source: "ContentView")
+
+        // Fire the duck first, then the start sound — at the very top of this
+        // function so they happen alongside the hotkey press, not 80ms later
+        // behind captureRecordingContext + setActiveRecordingMode +
+        // setOverlayMode. Duck-before-sound matters because CoreAudio's sound
+        // playback latency (~30-50ms) is shorter than the time it takes the
+        // fade to drop the music to a perceptibly quieter level; if the
+        // sound went first, you'd hear it before the music dipped.
+        let willStart = !self.asr.isRunning
+        var preAppliedMediaAction: MediaSessionAction = .none
+        if willStart, SettingsStore.shared.mediaBehaviorDuringTranscription == .duck {
+            if let prev = MediaPlaybackService.shared.duckSystemVolume() {
+                preAppliedMediaAction = .ducked(previousVolume: prev)
+            }
+        }
+        if willStart, SettingsStore.shared.enableTranscriptionSounds {
+            TranscriptionSoundPlayer.shared.playStartSound()
+        }
+
         self.captureRecordingContext()
         self.applyDictationShortcutSelectionContext(for: slot)
         self.setActiveRecordingMode(mode)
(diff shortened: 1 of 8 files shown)


altic-dev:
@AndrewBeniston  Thanks for the PR - seem useful to me :) But only ask would be to fix the Codex suggestions + Follow the PR template We have to help me review better. Also please attach some demo / pictures to show the working. Thank you!

PR #370, opened 2026-06-05
Add SenseVoice speech model option

jayrodge:
## Summary
- rework SenseVoice to use FluidAudio/CoreML instead of the sherpa-onnx-offline CLI
- keep SenseVoice English-only in FluidVoice and run the int8 CoreML encoder through FluidAudio on Apple Silicon
- remove the external CLI requirement, temporary WAV writes for sample transcription, ONNX archive download/extract flow, and sherpa setup docs
- add a focused FluidAudio companion branch dependency with SenseVoice CoreML support

## Companion dependency PR
- FluidAudio: https://github.com/altic-dev/FluidAudio/pull/1
- FluidVoice currently points at jayrodge/FluidAudio:add-sensevoice-coreml-for-fluidvoice so this PR can build before the companion FluidAudio branch is merged upstream.

## Verification
- swift build in the FluidAudio companion branch
- swift package resolve in FluidVoice
- xcodebuild -project Fluid.xcodeproj -scheme Fluid -configuration Debug -derivedDataPath /private/tmp/fluidvoice-derived CODE_SIGNING_ALLOWED=NO build -quiet

## Notes
- SenseVoice remains a batch/non-autoregressive model here, not a true live streaming engine.
- The implementation is now CoreML via FluidAudio, not ONNX via an external process.

### DIFF
--- a/Fluid.xcodeproj/project.pbxproj
+++ b/Fluid.xcodeproj/project.pbxproj
@@ -589,9 +589,9 @@
 /* Begin XCRemoteSwiftPackageReference section */
 		7C078DA32E3B386A00FB7CAC /* XCRemoteSwiftPackageReference "FluidAudio" */ = {
 			isa = XCRemoteSwiftPackageReference;
-			repositoryURL = "https://github.com/altic-dev/FluidAudio.git";
+			repositoryURL = "https://github.com/jayrodge/FluidAudio.git";
 			requirement = {
-				branch = "B/cohere-coreml-asr";
+				branch = "add-sensevoice-coreml-for-fluidvoice";
 				kind = branch;
 			};
 		};
--- a/Fluid.xcodeproj/project.pbxproj
+++ b/Fluid.xcodeproj/project.pbxproj
@@ -608,7 +608,7 @@
 			repositoryURL = "https://github.com/modelcontextprotocol/swift-sdk.git";
 			requirement = {
 				kind = upToNextMajorVersion;
-				minimumVersion = 0.10.2;
+				minimumVersion = 0.12.1;
 			};
 		};
 		7C3697872ED70F9C005874CE /* XCRemoteSwiftPackageReference "DynamicNotchKit" */ = {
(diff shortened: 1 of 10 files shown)


altic-dev:
Hi ! Thanks for the PR - few questions, just to make sure - this runs by calling a CLI under the hood to make the transcription?  FluidVoice itself doesn't handle the process right? 

- Does live preview work? 
- what's the underlying model format that is being executed - ONNX? 

though it's nice to have a new model, I need to figure out if CLI is a good way to integrate new models and what's the cons of it. Pro being easy integration of course. 

@grohith327  wdyt? 

Also, would appreciate if you add some screenshots or clips of it working :) 

Thanks!

grohith327:
Thanks for the PR @jayrodge, I see its still in draft but nevertheless will add my thoughts. I don't think CLI is good way to invoke the model as its calling the inference runtime (sherpa-onnx) outside of the app entirely. 

What happens to users who don't have it installed ? Don't want to just throw an error and leave it at that. What happens if the CLI freezes or crashes, will we be able to parse out related errors so users can report back ? 

Every transcription does a file write and read (unnecessary IO if you ask me). 

Looks like there are coreML version of this model and already supported through FluidInference. So that would be the ideal path forward for adding this model.

jayrodge:
Thanks for the feedback. I reworked this away from the CLI approach.

Current state:
- No more sherpa-onnx-offline process call.
- No ONNX runtime path in FluidVoice.
- No temporary WAV write for normal sample transcription.
- SenseVoice now calls FluidAudio's CoreML SenseVoiceManager directly.
- The model format is CoreML: FluidAudio downloads FluidInference/sensevoice-small-coreml and loads the int8 encoder variant.
- Live preview is not advertised as true streaming; SenseVoice stays batch/non-autoregressive in FluidVoice.

I opened a focused companion FluidAudio PR here: https://github.com/altic-dev/FluidAudio/pull/1

That branch is based on the FluidAudio branch FluidVoice already used and ports only the SenseVoice CoreML pieces needed here, so this avoids pulling in unrelated FluidAudio main-branch API churn. This FluidVoice PR currently points to my fork branch until that companion dependency path is agreed.

PR #379, opened 2026-06-12
Make the launch window optional for login-item launches

### LINKED ISSUE #369: [🐞 BUG] Preferences Window Opens on Startup When "Launch at Startup" Is Enabled
**Describe the bug**

When the Launch at startup option is enabled, Fluidvoice 1.5.15 launches the Preferences window automatically when macOS starts.

The application itself appears to start successfully, but instead of running silently in the background with only the tray/menu bar icon visible, the Preferences window is displayed every time the system starts.

**To Reproduce**

1. Open Fluidvoice 1.5.15.
2. Enable the Launch at startup option.
3. Quit Fluidvoice.
4. Restart macOS.
5. Log in to your account.
6. Observe that the Fluidvoice Preferences window opens automatically.

**Expected behavior**

When Launch at startup is enabled, Fluidvoice should start silently in the background and display only the tray/menu bar icon, remaining ready for use without opening the Preferences window.

**Environment:**
  
 - macOS Version: Tahoe 26.5.1
 - App Version: 1.5.15
 - Architecture: Apple Silicon

postoso:
Fixes #369

Per your comment on the issue — the window-on-login behavior was requested by users, so it stays the **default**. This adds a "Show window when launched at login" toggle (App Settings, right under "Launch at startup") for users who want FluidVoice to start silently in the menu bar.

## How it works
- Login-item launches are detected via the launch Apple Event (`kAELaunchedAsLogInItem`), read during `applicationDidFinishLaunching`. Manual launches always show the window regardless of the toggle, so double-clicking the app never feels broken.
- When the toggle is off, the main window is still **realized invisibly** (alpha 0, ordered out after startup completes) rather than skipped — ContentView's appearance bootstraps the menu bar and services, so the window must be created for the app to function on hidden launches. The existing reveal-retry loop and LaunchServices reopen fallback are reused; the fallback no longer activates the app for silent launches.
- `FLUID_SIMULATE_LOGIN_LAUNCH=1` env hook makes the login path testable without logging out.

## Tested
- Login launch + toggle ON (default): window shows — current behavior preserved
- Login launch + toggle OFF: no window, no focus steal; logs confirm ContentView startup ran hidden (menu bar + services boot normally); window opens fine afterward from the menu bar
- Manual launch + toggle OFF: window shows normally
- SwiftLint strict passes; no new warnings

### DIFF
--- a/Sources/Fluid/AppDelegate.swift
+++ b/Sources/Fluid/AppDelegate.swift
@@ -15,6 +15,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
     private var updateCheckTimer: Timer?
     private var didRevealMainWindowOnLaunch = false
     private var didRequestMainWindowReopen = false
+    private var shouldSuppressNextReopenActivation = false
     private var wasLaunchedAsLoginItem = false
 
     func applicationDidFinishLaunching(_ notification: Notification) {
@@ -74,6 +75,11 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
     }
 
     func applicationShouldHandleReopen(_ sender: NSApplication, hasVisibleWindows flag: Bool) -> Bool {
+        if self.shouldSuppressNextReopenActivation {
+            self.shouldSuppressNextReopenActivation = false
+            return true
+        }
+
         // Ensure dock-icon reopen always foregrounds FluidVoice.
         sender.activate(ignoringOtherApps: true)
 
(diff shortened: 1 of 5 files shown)


altic-dev:
down to merge it once you take care of codex' comments :) Thanks for the work @postoso !

postoso:
Thanks for jumping in and knocking out the Codex comments yourself, @altic-dev — appreciate it! 🙏 Looks good on my end, happy for it to go in whenever you're ready. I'll watch the Codex feedback on #378 (diarization) and handle those directly.

PR #384, opened 2026-06-19
fix: surface Apple speech models in onboarding (#354)

### LINKED ISSUE #354: [🐞 BUG] Onboarding has no offline path — zero-download Apple speech models aren't selectable in setup
**Describe the bug**
The onboarding "voice model" step (Step 1) offers **only models that require a download** from HuggingFace (Parakeet / Cohere / Whisper). The built-in, **zero-download Apple speech models** (`apple-speech` "Apple ASR Legacy", `apple-speech-analyzer` "Apple Speech - macOS 26+") are **not selectable during onboarding** — they only appear later in Settings.

As a result, a user on a restricted network where the model download is blocked has **no way to complete onboarding at all** — every offered option fails to download, and the only models that need no network are hidden.

**Where**
`Sources/Fluid/UI/WelcomeView.swift` — the onboarding candidate list is hardcoded:
```swift
private var onboardingModelOptions: [SettingsStore.SpeechModel] {
    let candidates: [SettingsStore.SpeechModel] = CPUArchitecture.isAppleSilicon
        ? [.parakeetTDT, .cohereTranscribeSixBit, .parakeetRealtime, .parakeetTDTv2, .whisperBase, .whisperSmall]
        : [.whisperBase, .whisperTiny, .whisperSmall, .whisperMedium]
    ...
}
```
`appleSpeech` / `appleSpeechAnalyzer` are never in the candidate set, even though `SpeechModel.availableModels` includes them and they are marked "Built-in (Zero Download)".

**Impact**
- Users behind a firewall/proxy are completely stranded at setup.
- Ironically, the Apple on-device models are often exactly what privacy/policy-restricted users need.

**Suggested fix**
- Include the available Apple speech model(s) as an onboarding option (e.g. an "Use built-in Apple speech (no download)" choice), gated by OS/arch availability as already done in `SpeechModel.availableModels`.
- Ideally surface it prominently when a download fails, as an offline fallback.

**Workaround (for reference)**
Set defaults manually to bypass: `SelectedSpeechModel = "apple-speech-analyzer"` and `OnboardingCompleted = true`, then relaunch.

**Environment**
- macOS 26 (Tahoe), Apple Silicon

postoso:
## What
Adds the built-in Apple speech models (Apple ASR Legacy, Apple Speech for macOS 26+) to the onboarding voice-model picker. Previously onboarding only offered download-required models (Parakeet/Cohere/Whisper), so a user on a restricted network where the HuggingFace download is blocked could not complete setup.

## Why
The two Apple models were absent from the hardcoded onboarding candidate list even though `SpeechModel.availableModels` already admits them. `apple-speech` runs fully on-device with no download — the true offline fallback. `apple-speech-analyzer` is an on-device OS API (macOS 26+) but may fetch locale speech assets on first use, so it is not guaranteed zero-download; both are valid options, gated by OS/arch.

## How
- Add `.appleSpeech` + `.appleSpeechAnalyzer` (before the Whisper entries) to both candidate branches; the existing `availableModels` filter gates them by OS/arch (analyzer excluded < macOS 26).
- Built-ins show a no-download "Use" button and selection proceeds with no network.
- Update the "Other" language subtitle to name the no-download Apple Speech option.

## Notes
On Apple Silicon the Apple models surface under "More options" (via the "Other" language path). Surfacing them in the default flow / prominently on a download failure is a possible follow-up.

## Verification
`swiftlint --strict` clean; `xcodebuild build` succeeds.

Closes #354

### DIFF
--- a/Sources/Fluid/UI/WelcomeView.swift
+++ b/Sources/Fluid/UI/WelcomeView.swift
@@ -1044,7 +1044,12 @@ struct OnboardingFlowView: View {
                                 .foregroundStyle(self.theme.palette.primaryText)
 
                             VStack(spacing: 8) {
-                                ForEach(Array(self.onboardingAlternativeModels.prefix(3))) { model in
+                                // Cap of 5 fits the longest alternatives list any branch
+                                // produces (Intel + macOS 26: Apple Speech, Apple Speech
+                                // Analyzer, Whisper Tiny/Small/Medium), so adding the Apple
+                                // offline models never evicts a previously reachable Whisper
+                                // option.
+                                ForEach(Array(self.onboardingAlternativeModels.prefix(5))) { model in
                                     self.onboardingModelOptionRow(for: model)
                                 }
                             }
--- a/Sources/Fluid/UI/WelcomeView.swift
+++ b/Sources/Fluid/UI/WelcomeView.swift
@@ -594,7 +594,7 @@ struct OnboardingFlowView: View {
             case .multipleLanguages:
                 return "Uses Parakeet TDT v3 or Cohere"
             case .other:
-                return "Whisper and manual choices"
+                return "Whisper, Apple Speech (no download), and manual choices"
             }
         }
     }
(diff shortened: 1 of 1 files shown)


postoso:
Superseded by 1.6.0. The new Apple Speech voice engine option with no-download, language-aware setup covers what this PR was doing for #354 (making the zero-download Apple speech models selectable during onboarding). Closing in favor of the shipped version. Thanks for getting it in.

PR #385, opened 2026-06-19
fix: guard against NSNotFound selection-range overflow crashes (#319)

### LINKED ISSUE #319: [🐞 BUG] Repeatable crash in TypingService.PasteboardRestore queue (EXC_BREAKPOINT brk #1) — affects both insertion modes via fallback
## Summary

FluidVoice 1.5.13 crashes deterministically with `EXC_BREAKPOINT` (`brk #1`, Swift fatal-error trap) on the `TypingService.PasteboardRestore` dispatch queue. I've reproduced it 5 times across 3 sessions, all identical signatures.

The same crash hits **both** `Clipboard Paste` and `Clipboard Free Insert` modes, because the keystroke path falls back to `insertTextViaClipboard` at `Sources/Fluid/Services/TypingService.swift:368` when CGEvent + Accessibility both fail — so toggling the setting is not a workaround.

## Environment

- FluidVoice **1.5.13** (build 10), installed from official `.dmg`
- macOS 26.3.1 (25D771280a)
- Mac14,7 (Apple Silicon)

## Crash signature (5/5 reports identical)

```
exception : EXC_BREAKPOINT (SIGTRAP)
ESR       : (Breakpoint) brk 1
queue     : TypingService.PasteboardRestore
```

Faulting frames (binary is stripped, so only offsets resolve):

```
0x100124b78 (in FluidVoice) + 1576    ← PC, brk #1
0x100124390 (in FluidVoice) + 64
0x100105bb8 (in FluidVoice) + 28
_dispatch_call_block_and_release
_dispatch_lane_serial_drain
...
```

`+1576` places the trap deep inside a substantial function. By elimination via code audit (below), that function is almost certainly `waitForFocusedTextVerification`.

## Audit — what runs on `pasteboardRestoreQueue`

The closure at `TypingService.swift:553` only calls:

- `waitForFocusedTextVerification`
- `NSPasteboard` accessors (cannot `brk #1`)
- `restorePasteboardSnapshot` (no force-unwraps, no subscripts)
- `log`

`brk #1` candidates inside `waitForFocusedTextVerification` (lines ~937–1000):

| Line(s) | Code | Risk |
|---|---|---|
| 973, 991 | `before.location + expectedLength` | **High** — overflows if AX returns `NSNotFound` (= `Int.max`) for selected-range location |
| 974, 992 | `abs(after.location - expectedCaretLocation)` | High — same root cause |

No force-unwraps, `as!` casts, or array subscripts on this path. The **integer-overflow** hypothesis is the only viable explanation in the file.

## Why `Clipboard Free Insert` doesn't avoid it

`TypingService.swift:368` — when CGEvent + Accessibility both fail (common against stubborn Electron / GPU-terminal / web targets), the pipeline falls through to `insertTextViaClipboard`, which dispatches onto `pasteboardRestoreQueue` and re-enters the same trap.

```swift
// Fallback: Use clipboard-based insertion (more reliable)
self.log("[TypingService] CGEvent failed, trying clipboard fallback")
if self.insertTextViaClipboard(text) { ... }   // ← re-enters the buggy queue
```

## Suggested fix

Guard the arithmetic in `waitForFocusedTextVerification` against `NSNotFound`/overflow before using `before.location`:

```swift
// In both the `appScriptSelectedRange` and `selectedRange` branches:
guard before.location != NSNotFound,
      after.location  != NSNotFound,
      after.length == 0
else { /* skip this signal, keep polling */ }

let (expectedCaretLocation, addOverflow) =
    before.location.addingReportingOverflow(expectedLength)
guard !addOverflow else { continue }

let (rawDelta, subOverflow) =
    after.location.subtractingReportingOverflow(expectedCaretLocation)
guard !subOverflow else { continue }
let caretDelta = abs(rawDelta)
```

A defense-in-depth alternative is to sanitize at the source in `getSelectedTextRange` — return `nil` if `range.location == NSNotFound` or `range.location < 0`. That also protects any future callers.

## What I'd need to be 100% sure of the line

The shipped binary is fully stripped (`nm` shows imported symbols only), and no dSYM is published with the v1.5.13 release. The atos output above is the best I can recover from outside.

Could you either:

1. Attach `Fluid-oss-1.5.13.dSYM.zip` to the v1.5.13 GitHub release, or
2. Paste the output of:
   ```
   atos -arch arm64 -o /path/to/FluidVoice.app/Contents/MacOS/FluidVoice -l <load-addr> 0x101000b78
   ```
   from a build that has symbols.

That pins the exact line, and I'll send a PR.

## Crash report

Available on request — `.ips` files contain hardware identifiers I'd rather not paste publicly. Happy to share via a private channel.

postoso:
## What
Fixes a reproducible crash (EXC_BREAKPOINT / Swift integer-overflow trap) ~5 s after a successful clipboard-mode text insertion ("one good dictation per launch, then a crash").

## Root cause
macOS 26 AX APIs can report a selected-text range of `{location: NSNotFound (Int.max), length: 0}`. Feeding that raw location into caret/bounds arithmetic (`before.location + expectedLength`; `range.location + range.length`) overflows and traps.

## Fix (defense-in-depth)
- `TypingService.getSelectedTextRange` sanitizes the range (rejects `NSNotFound`/negative).
- The caret-distance check is extracted into an overflow-safe helper (`addingReportingOverflow` / `.magnitude`), preserving the original tolerance semantics.
- `TextSelectionService` guard now catches `NSNotFound` (not just `kCFNotFound == -1`), and its bounds validation is overflow-safe via `boundedSelectionRange`.
- 10 unit tests covering the sanitizer, caret-distance, and bounds helpers.

## Verification
`swiftlint --strict` clean; `xcodebuild build` succeeds; the 10 unit tests pass.

Closes #319

### DIFF
--- a/Sources/Fluid/Services/TextSelectionService.swift
+++ b/Sources/Fluid/Services/TextSelectionService.swift
@@ -108,7 +108,7 @@ final class TextSelectionService {
             return nil
         }
 
-        guard range.location != kCFNotFound, range.length > 0 else {
+        guard range.location >= 0, range.location != NSNotFound, range.length > 0 else {
             self.diag("Selected range empty (location=\(range.location), length=\(range.length))")
             return nil
         }
@@ -121,17 +121,26 @@ final class TextSelectionService {
         }
 
         let nsText = fullText as NSString
-        guard range.location >= 0,
-              range.length > 0,
-              range.location + range.length <= nsText.length
-        else {
+        guard let safeRange = Self.boundedSelectionRange(range, textLength: nsText.length) else {
             self.diag("Selected range out of bounds (textLen=\(nsText.length), location=\(range.location), length=\(range.length))")
             return nil
         }
 
-        let extracted = nsText.substring(with: NSRange(location: range.location, length: range.length))
+        let extracted = nsText.substring(with: safeRange)
         self.diag("Selected range extraction succeeded (chars=\(extracted.count))")
         return extracted
+    }
+
+    /// Overflow-safe bounds validation for an AX-derived selection range against `textLength`.
+    /// Returns the in-bounds `NSRange`, or nil when the range is invalid or out of bounds.
+    /// macOS 26 can report `kAXSelectedTextRangeAttribute` with `location == NSNotFound` (`Int.max`);
+    /// using `addingReportingOverflow` here prevents the `location + length` integer-overflow trap
+    /// such a value would otherwise cause (the same #319 crash class as TypingService).
+    static func boundedSelectionRange(_ range: CFRange, textLength: Int) -> NSRange? {
+        guard range.location >= 0, range.location != NSNotFound, range.length > 0 else { return nil }
+        let (rangeEnd, overflowed) = range.location.addingReportingOverflow(range.length)
+        guard !overflowed, rangeEnd <= textLength else { return nil }
+        return NSRange(location: range.location, length: range.length)
     }
 
     private func describe(_ error: AXError) -> String {
--- a/Sources/Fluid/Services/TypingService.swift
+++ b/Sources/Fluid/Services/TypingService.swift
@@ -981,18 +981,13 @@ final class TypingService {
 
         var range = CFRange()
         let ok = AXValueGetValue(unsafeBitCast(axValue, to: AXValue.self), .cfRange, &range)
-        return ok ? Self.sanitizedSelectedTextRange(range) : nil
-    }
-
-    /// Rejects selection ranges that macOS reports when there is no valid caret/selection.
-    /// On macOS 26, `kAXSelectedTextRangeAttribute` can come back as `{location: NSNotFound, length: 0}`
-    /// (`NSNotFound == Int.max`); feeding that raw location into caret arithmetic
-    /// (`before.location + expectedLength`) overflows and traps. Returning nil here keeps both
-    /// callers (`captureFocusedTextSnapshot`, `insertTextAtCursorUsingSelectedRange`) on their
-    /// existing nil-handling paths instead of operating on a poisoned range.
-    static func sanitizedSelectedTextRange(_ range: CFRange) -> CFRange? {
-        guard range.location >= 0, range.location != NSNotFound, range.length >= 0 else { return nil }
-        return range
+        // Return the raw range. macOS 26 can report a sentinel `{location: NSNotFound, length: 0}`,
+        // but both callers handle it safely: `insertTextAtCursorUsingSelectedRange` clamps the
+        // location to the field length (inserting at the end, preserving contents), and the deferred
+        // verification path runs it through the overflow-safe `caretMovedExpectedDistance`. Rejecting
+        // the sentinel in this shared getter instead caused approach-0 insertion to bail to a
+        // whole-field overwrite (#319 regression), so sanitization stays out of the getter.
+        return ok ? range : nil
     }
 
     private func captureFocusedTextSnapshot() -> FocusedTextSnapshot? {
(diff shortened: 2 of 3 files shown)


altic-dev:
If you can provide a screencapture of the crash, that would be really helpful, like a video clip or something. Because I'm not able to reproduce it, so I don't know if this fix is gonna cause more issues, if that makes sense. Other than that, I think it's a beautiful PR, would love to merge this, if you can prove me that it's a fix for a reproducible issue xD

postoso:
Hey @altic-dev, fair ask. This PR fixes #319, which @domci reported with a full crash signature (`EXC_BREAKPOINT` / `brk #1` on the `TypingService.PasteboardRestore` queue) and offered to share the `.ips` for. I jumped in as the fixer after digging through the code, so I don't actually have a recording of it myself. The good news is it's deterministic, not a heisenbug.

Here's what happens. macOS AX sometimes hands back a selected-text range of `{location: NSNotFound, length: 0}`, and `NSNotFound` is just `Int.max`. That goes straight into the caret math (`before.location + expectedLength`, still sitting at `TypingService.swift:1067`/`1085` on main), overflows, and traps. That's your `brk #1`. It only shows up when CGEvent + AX insertion both fail and the code falls back to the clipboard path (Electron apps, GPU terminals, some web fields), so a normal text field never hits it. That's almost certainly why it wouldn't repro for you.

The 10 unit tests feed that exact `NSNotFound` range into the real arithmetic and check it traps before the fix and stays safe after, so even without filming the live crash the fix is provably doing its job. And it's pure defense-in-depth: nothing changes on the happy path, it just won't run overflowing math on a "no selection" sentinel.

@domci, you've got the setup that reproduces it. Any chance you could drop the crash log or a quick clip on #319? That'd give @altic-dev the end-to-end proof.

Rebased onto 1.6.0; builds clean, swiftlint --strict clean, and all 10 unit tests pass against current main.

domci:
> Hey @altic-dev, fair ask. This PR fixes #319, which @domci reported with a full crash signature (`EXC_BREAKPOINT` / `brk #1` on the `TypingService.PasteboardRestore` queue) and offered to share the `.ips` for. I jumped in as the fixer after digging through the code, so I don't actually have a recording of it myself. The good news is it's deterministic, not a heisenbug.
> 
> Here's what happens. macOS AX sometimes hands back a selected-text range of `{location: NSNotFound, length: 0}`, and `NSNotFound` is just `Int.max`. That goes straight into the caret math (`before.location + expectedLength`, still sitting at `TypingService.swift:1067`/`1085` on main), overflows, and traps. That's your `brk #1`. It only shows up when CGEvent + AX insertion both fail and the code falls back to the clipboard path (Electron apps, GPU terminals, some web fields), so a normal text field never hits it. That's almost certainly why it wouldn't repro for you.
> 
> The 10 unit tests feed that exact `NSNotFound` range into the real arithmetic and check it traps before the fix and stays safe after, so even without filming the live crash the fix is provably doing its job. And it's pure defense-in-depth: nothing changes on the happy path, it just won't run overflowing math on a "no selection" sentinel.
> 
> @domci, you've got the setup that reproduces it. Any chance you could drop the crash log or a quick clip on #319? That'd give @altic-dev the end-to-end proof.
> 
> Rebased onto 1.6.0; builds clean, swiftlint --strict clean, and all 10 unit tests pass against current main.

hi @postoso i sure would love to. but im on Version 1.6.0 (12) now and the bug seems gone. I cant reproduce anymore.

postoso:
Thanks for checking, @domci, glad it's not hitting you on 1.6.0(12) anymore.

To be straight about what this is now: defense-in-depth, not a fix for a live repro. The NSNotFound / Int.max range still feeds the caret math on the clipboard-fallback path, and the 10 unit tests prove that arithmetic traps before this change and stays safe after, so the guard holds whether or not the crash currently reproduces, and the happy path is untouched.

@altic-dev, your call on whether that's worth keeping. Glad to leave it in as a latent-overflow guard, or close it if you'd rather not carry a fix for something that's not currently reproducible. Either works for me.

postoso:
Closing this to reduce the stale PR queue. The original reporter can no longer reproduce the crash on 1.6.0, and without a current repro or clip this is now only a defense-in-depth guard rather than a proved live fix. Happy to revisit if the crash comes back with a fresh log or reproduction.

PR #386, opened 2026-06-19
fix: validate downloaded model content + surface onboarding errors (#353, #355)

### LINKED ISSUE #353: [🐞 BUG] ASR model downloader saves HTML proxy/error pages as model files (no content validation)
**Describe the bug**
The ASR model downloader does not validate the *content* of what it downloads. On a network behind a corporate web proxy/SWG, the proxy can return an **HTML block/notification page with HTTP 200** instead of the real file. The downloader writes that HTML straight to disk as the model payload, so every `coremldata.bin` ends up being `<!DOCTYPE html>…` rather than a CoreML binary.

The app then reports the model as cached and tries to load it, and CoreML fails with an opaque error.

**Evidence (real failure)**
Loading error:
```
[ERROR] [ASRService] ... Unable to load model ... parakeet-tdt-0.6b-v2-coreml/Preprocessor.mlmodelc/coremldata.bin
is not a valid .mlmodelc file because the first word (1329865020) is not recognizable.
```
`1329865020` decoded little-endian begins with the bytes `3C 21` = `<!` — i.e. the file starts with `<!DOCTYPE html>`.

Inspecting the cache (`~/Library/Application Support/FluidAudio/Models/...`):
```
$ file .../Preprocessor.mlmodelc/coremldata.bin
.../coremldata.bin: HTML document text, UTF-8 text, with very long lines

$ head -c 80 .../coremldata.bin
<!DOCTYPE html><html lang="en"><head><meta charSet="utf-8"/>...
```
**Every** `coremldata.bin` across all model variants (v2, v2-coreml, v3) was the identical ~22 KB HTML page — clearly a proxy-injected block page returned with HTTP 200, not the model.

**Impact**
- Silent corruption: the app believes models are "cached on disk" and never recovers.
- Particularly affects anyone on a corporate/enterprise network.

**Suggested fix**
In the model download path, validate each downloaded artifact before persisting:
- Check HTTP `Content-Type` (reject `text/html`).
- Check magic bytes / expected size of `coremldata.bin` (and/or the `.tar`), reject obviously-wrong payloads.
- Treat redirects to non-CDN hosts as suspect.
- On validation failure, delete the partial file and surface a clear, actionable error (see related issue on missing onboarding error feedback).

Note: the download likely lives in the FluidAudio dependency (`AsrModels.downloadAndLoad`), so the fix may belong upstream there — but FluidVoice could also validate post-download.

**Environment**
- macOS 26 (Tahoe), Apple Silicon
- Network: behind a corporate web proxy that returns an HTML page for blocked/intercepted downloads

### LINKED ISSUE #355: [🐞 BUG] No UI error feedback when ASR model download/load fails during onboarding
**Describe the bug**
When an ASR model fails to download or load during onboarding, the UI gives **no feedback about what went wrong**. The setup screen simply doesn't advance, with no error message, retry guidance, or indication that the download was blocked/failed. The failure is only visible in the log file.

This compounds two other problems (download content not validated; no offline model option in onboarding): the user is stuck at "select a model to download" with no way forward and no explanation.

**Evidence**
The failure is logged but never shown to the user:
```
[ERROR] [ASRService] ASRService: First prepare attempt for FluidAudio ... failed after 5.77s
[WARN]  [ASRService] ... Unable to load model ... coremldata.bin is not a valid .mlmodelc file ...
```
Meanwhile the onboarding UI shows no error state and the "Continue/Finish Setup" gate stays disabled.

**Suggested fix**
- Surface download/load failures in the onboarding UI with a clear message and cause (e.g. "Download failed / blocked", "Downloaded file was not a valid model", network/proxy hint).
- Offer actionable next steps: retry, choose a different (or built-in Apple) model, or manual/offline model staging instructions.
- Consider a visible error state on the model card rather than a silently-disabled Continue button.

**Environment**
- macOS 26 (Tahoe), Apple Silicon
- Network: model download blocked/intercepted by a corporate proxy

postoso:
## What
Two related onboarding/download robustness fixes.

### #353 — reject HTML/markup pages saved as model files
A network proxy / secure web gateway can return an HTML block page with HTTP 200 in place of the real file. The downloader persisted that page as the model (e.g. `coremldata.bin`), permanently caching a corrupt model that never self-heals.

- `validateDownloadedFile` now runs before the temp file is moved into place. It rejects markup by `Content-Type` (`text/html`, `text/xml`, `application/xml`) **and** by byte-sniffing the leading 512 bytes: after stripping a UTF-8 BOM + ASCII whitespace, any payload beginning with `<` followed by a markup-ish byte (ASCII letter, `!`, `?`, `/`) is rejected. No artifact this downloader fetches begins with `<` (CoreML binaries; JSON `{`/`[`; `model.mil` → `program`), so false-positive risk is negligible.
- A rejected/partial temp file is now removed instead of cached.

### #355 — surface onboarding model-load failures
Onboarding load/download failures previously only logged, leaving the user stuck at a disabled Continue button with no message. The onboarding `catch` now sets the existing `asr` error state and the onboarding view presents a Retry/OK alert.

## Verification
`swiftlint --strict` clean; `xcodebuild build` succeeds; 4 new unit tests for the markup detection pass (rejects `<!doctype`/`<html`/`<script`/`<head`/`<meta`/`<!--`/`<?xml`, incl. BOM + leading-whitespace variants; accepts JSON `{`/`[`, `.mil`, and binary/CoreML-magic prefixes).

## Note
The Parakeet path fetches models through the upstream FluidAudio package (`AsrModels.downloadAndLoad`), which doesn't go through this downloader; this hardening covers the HuggingFace / Whisper / Nemotron / external-CoreML path. A FluidAudio-side guard would be a separate upstream change.

Closes #353
Closes #355

### DIFF
--- a/Sources/Fluid/Networking/ModelDownloader.swift
+++ b/Sources/Fluid/Networking/ModelDownloader.swift
@@ -300,6 +300,51 @@ final class HuggingFaceModelDownloader {
         return Self.looksLikeHTML(prefix)
     }
 
+    /// Returns `true` if any cached payload under `relativePaths` (each resolved against `root`)
+    /// is an HTML/markup document rather than real model data — the cached-*tree* analog of
+    /// `cachedFileIsMarkup`, intended for a provider preflight to call before trusting a present
+    /// cache and skipping the downloader. The downloader itself already re-validates each file via
+    /// `needsDownload`, but a preflight that returns on file-existence alone never reaches it, so a
+    /// corrupt-but-present cache would slip through (see #353).
+    ///
+    /// Each relative path may be a regular file or a directory (e.g. a `.mlpackage` bundle).
+    /// Directories are scanned recursively and every regular file inside is byte-sniffed with
+    /// `cachedFileIsMarkup`, reusing the single `looksLikeHTML` detector — there is no second
+    /// markup heuristic. Conservative on uncertainty, mirroring `cachedFileIsMarkup`: a path that
+    /// does not exist, a file that cannot be read, or a directory that cannot be enumerated is
+    /// skipped (treated as non-markup), so a valid cache is never reported corrupt. An empty
+    /// required directory therefore yields `false` here — its incompleteness is the existence
+    /// check's concern, not this markup check's.
+    static func cachedPayloadContainsMarkup(root: URL, relativePaths: [String]) -> Bool {
+        let fileManager = FileManager.default
+        for relativePath in relativePaths {
+            let url = root.appendingPathComponent(relativePath)
+            var isDirectory: ObjCBool = false
+            guard fileManager.fileExists(atPath: url.path, isDirectory: &isDirectory) else {
+                continue
+            }
+            if isDirectory.boolValue {
+                guard let enumerator = fileManager.enumerator(
+                    at: url,
+                    includingPropertiesForKeys: [.isRegularFileKey],
+                    options: [.skipsHiddenFiles]
+                ) else {
+                    continue
+                }
+                for case let fileURL as URL in enumerator {
+                    let isRegularFile = (try? fileURL.resourceValues(forKeys: [.isRegularFileKey]))?.isRegularFile ?? false
+                    guard isRegularFile else { continue }
+                    if Self.cachedFileIsMarkup(at: fileURL) {
+                        return true
+                    }
+                }
+            } else if Self.cachedFileIsMarkup(at: url) {
+                return true
+            }
+        }
+        return false
+    }
+
     /// Returns `true` if `data` begins with an HTML / XML markup marker, ignoring a leading
     /// UTF-8 BOM and ASCII whitespace.
     ///
--- a/Sources/Fluid/Networking/ModelDownloader.swift
+++ b/Sources/Fluid/Networking/ModelDownloader.swift
@@ -196,6 +196,10 @@ final class HuggingFaceModelDownloader {
                         continuation.resume(throwing: NSError(domain: "HF", code: http.statusCode))
                         return
                     }
+                    // Reject HTML error/block pages (e.g. a corporate proxy returning its
+                    // notification page with HTTP 200) before persisting them as a model
+                    // file, otherwise a corrupt payload is cached permanently. See #353.
+                    try Self.validateDownloadedFile(at: tempUrl, response: response, relativePath: relativePath)
                     try FileManager.default.createDirectory(at: destination.deletingLastPathComponent(), withIntermediateDirectories: true)
                     if FileManager.default.fileExists(atPath: destination.path) {
                         try FileManager.default.removeItem(at: destination)
@@ -203,6 +207,8 @@ final class HuggingFaceModelDownloader {
                     try FileManager.default.moveItem(at: tempUrl, to: destination)
                     continuation.resume()
                 } catch {
+                    // Never leave a rejected/partial payload behind.
+                    try? FileManager.default.removeItem(at: tempUrl)
                     continuation.resume(throwing: error)
                 }
             }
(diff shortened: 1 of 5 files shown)


altic-dev:
Hey man. Thanks for the PRs! I am working on a huge push and it might conflict with your PRs. Would appreciate if you can hold on for a week or lesser until I push 1.6.0 and we can work on pulling your changes in :) wdyt?

postoso:
Sounds great — totally happy to hold. 🙌 Go knock out 1.6.0, no rush on my end. Once it lands I'll rebase my open PRs (#384, #385, #387, #386, and #378/diarization) on top so they're clean for you to pull in. Just give me a ping when it's out. 🙏

postoso:
Rebased onto 1.6.0. The #353 downloader validation applied cleanly. The #355 onboarding-error piece needed rework: 1.6 rewrote onboarding and renamed prepareOnboardingModel to prepareOnboardingRoute, so the original standalone .alert (its Retry called the old name) no longer compiled. Turns out 1.6's ContentView alert is bound to asr.showError and already wraps the onboarding view, and 1.6 already uses that pattern for delete failures but left download/load failures log-only. So I kept the real #355 fix (set errorTitle/errorMessage/showError in the prepare catch, so failures actually surface) and dropped the now-redundant standalone alert. Builds green, swiftlint clean. Flagging the dropped alert here in case you would rather onboarding own a dedicated Retry alert instead.

postoso:
@codex review

postoso:
Pushed a follow-up addressing the Codex review's cached-file point: `ensureModelsPresent` now re-validates files already on disk (reusing the same `looksLikeHTML` byte-sniff), so a corrupt HTML payload cached before this fix gets deleted and re-downloaded instead of staying stuck.

Codex's second point is fair too: `NemotronProvider` and `ExternalCoreML` can skip `ensureModelsPresent` via their own `modelsExistOnDisk()` checks, so a corrupt-but-present cache could still slip through at the provider preflight level. I left that out of this PR on purpose to keep it scoped to #353 (the downloader itself). Pushing the markup check down into each provider's presence logic is a broader change and a design call on where that validation should live, so it's probably better as a separate follow-up if you want it. Happy to do it.

postoso:
Update on the provider-level point from the last Codex review: I went ahead and closed it in this PR rather than splitting it out. `NemotronProvider` and `ExternalCoreML` now byte-sniff their cached binaries through the same shared check, so a corrupt-but-present cache gets re-downloaded at the preflight too, not just in the downloader. So #386 now fully closes #353 end to end: downloader content-validation + re-validating already-cached files + both provider preflights. Codex re-reviewed and it's clean.

altic-dev:
Thanks for a beautiful PR. merged :)

postoso:
Hey @altic-dev, thanks for merging #386, and congrats on 1.6.0. You'd mentioned holding my PRs until it shipped so we could pull them in after, so here's where the rest stand. All rebased onto 1.6.0 and building green, ready whenever you have a window:

- #398: honor hide-from-dock on login/cold launch (#396), the dock-icon-on-first-launch one
- #403: stop persisting dictated content to the disk log (privacy)
- #402: hide the dictation overlay from VoiceOver, fast-path system shortcuts (#401)
- #387: clipboard fallback for Write Mode selection reading (#259)
- #378: optional speaker diarization for file transcription (feature, no rush)

No pressure on timing, and happy to rebase again if your next push moves things. Thanks for building this.

altic-dev:
@postoso  sure! Thanks for the PRs - very helpful. But it will make it insanely helpful if you could follow the PR template that we got and make it easy for me to understand the issues and the fix. Currrent PRs are way too Claude slop and it's very hard for me to review it manually and that's why i am not able to merge it as soon as possible. 

- Follow template
- attach some video/screenshot of before and after or new. feature etc if possible 
- Codex review fixes ( This you're doing okay ) 

all of these. help me understand the issues and fixes,

The main reason is I need to make sure it doesn't break anything and works across thousands of computers out there Which makes it harder for me to just merge it when I think the Code seems okay, but there is a lot more verification I do internal data make sure it works pretty decently. Hope it makes sense.

altic-dev:
Another example  - https://github.com/altic-dev/FluidVoice/pull/378

an amazing feature add on which I'd really love and would merge it instantly if there's more than just 100 lines and a PR request :) Makes it very very hard for me to review it. 

"Is the code change here doing what it's supposed to do?" 
- To answer this - I need to know what you're trying to do. exact features and what do I see after compiling it. If I dont have both of these, then I need to work on it more to figure both and I still won't know what your intention was fully. 

I hope this makes it a little easier. to understand. :)

PR #392, opened 2026-06-21
Fix dictation prompt role separation

### LINKED ISSUE #388: [🐞 BUG] AI Request formed in a way that is much more likely to lead to failed instruction following (answering the transcript instead of cleaning it).
The AI prompts really need to put the user instructions in the `system prompt` NOT in the user message w/ the transcript payload appended directly after it. I was wondering why my prompt was having some relatively decent models fail miserably (like gpt-oss-120b w/ cerebras for insane speed) 

Dumb failure #1
Image

Dumb failure #2
Image

But then this works??? lol
Image

I would just use a different model, but gpt-oss-120b from cerebras really occupies a special place: It's cheap, it *can* be smart and clean up transcripts well, and most importantly it's insanely fast (up to ~1000 tokens/second

## CURRENT METHOD OF SENDING AI PROMPTS *using {{placeholders}}*:
```json
{
  "messages": [
    {
      "role": "user",
      "content": "{{INSTRUCTIONS}}\n\n{{TRANSCRIPT}}"
    }
  ],
  "model": "gpt-oss-120b",
  "reasoning_effort": "low"
}
```
>  the `\n\n` inst a clean enough seperation for some of the more persnickety models like gpt-oss-120b. This is a very important model for dictation cleanup though because it gets ~1000+ tokens/second w/ cerebras API. The model IS smart enough for the task, you just need to put the dictation cleanup instructions in the system prompt

## HOW IT SHOULD BE:
```json
{
  "messages": [
    {
      "role": "system",
      "content": "{{INSTRUCTIONS}}"
    }
    {
      "role": "user",
      "content": "<transcript>\n{{TRANSCRIPT}}\n</transcript>"
    }
  ],
  "model": "gpt-oss-120b",
  "reasoning_effort": "low"
}
```

## OR FOR THE GUIDED EDIT FEATURE:
```json
{
  "messages": [
    {
      "role": "system",
      "content": "{{INSTRUCTIONS}}"
    }
    {
      "role": "user",
      "content": "<verbal_instructions>\n{{VERBAL_INSTRUCTIONS}}\n</verbal_instructions>\n\n{{<context>\n{{SELECTED_TEXT}}\n</context>"
    }
  ],
  "model": "gpt-oss-120b",
  "reasoning_effort": "low"
}
```

- Some older local LLMs don't support system prompts, but that is the exception, not the rule, the app already kicks back bad requests when something like wrong reasoning-level chosen, so i don't think creating this edge case by using system prompt is a major issue, just provide a toggle in the AI instructions for "put instructions in user message instead of system prompt"; 

- OR better yet, make a **manual mode** where you just let the user form the entire json payload using `{{...}}` placeholders like {{SYSTEM_PROMPT}}, {{TRANSCRIPT}}, {{EDIT_INSTRUCTIONS}}, {{SELECTED_TEXT}} or %transcript%, %system_prompt%, etc.

None of the dictation apps do this which inevitebly leaves me at an impasse when i'm trying to work with a stubborn model like gpt-oss-120b that really needs so deliberate request structuring.

sdhilip200:
## Description
Fixes dictation post-processing prompt composition so cleanup instructions are sent as the system message and the transcript is sent separately as tagged user content. Legacy custom prompts that still use `${transcript}` continue to render with replacement behavior.

## Type of Change
- [ ] 📝 Documentation update

## Related Issues
- Closes #388

## Testing
- [ ] Tested on Intel Mac
- [x] Tested on macOS 26.2
- [ ] Built locally: `sh build_incremental.sh`

Additional validation run locally:
- [x] `xcodebuild clean build -project Fluid.xcodeproj -scheme Fluid -destination platform=macOS,arch=arm64 -derivedDataPath /private/tmp/FluidVoiceDerivedData CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO -quiet`
- [x] `xcodebuild test -project Fluid.xcodeproj -scheme Fluid -destination platform=macOS,arch=arm64 -derivedDataPath /private/tmp/FluidVoiceDerivedData CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO -quiet`
- [x] `git diff --check`

## Notes
- The SwiftPM lockfile updates `modelcontextprotocol/swift-sdk` to 0.12.1 because the previously pinned 0.10.2 fails to compile under Xcode 26.5 / Swift 6 strict-concurrency checking.
- `sh build_incremental.sh` currently fails in this checkout because it references `build_dev.sh`, which is not present in the repository. Direct `xcodebuild` build and test commands passed instead.
- Full `swiftformat --config .swiftformat Sources` was not committed because it rewrites broad pre-existing formatting drift outside this focused bugfix.

## Screenshots / Video 
N/A - no UI changes.

### DIFF
--- a/Fluid.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Fluid.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -83,6 +83,15 @@
       }
     },
     {
+      "identity" : "swift-atomics",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/apple/swift-atomics.git",
+      "state" : {
+        "revision" : "b601256eab081c0f92f059e12818ac1d4f178ff7",
+        "version" : "1.3.0"
+      }
+    },
+    {
       "identity" : "swift-collections",
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/swift-collections.git",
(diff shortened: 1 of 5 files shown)


sdhilip200:
Rebased onto current \ and resolved the conflict in \ by preserving the newer package pins alongside this change. Force-pushed the updated branch. I attempted \, but this machine is currently blocked by a local Swift toolchain/SDK mismatch plus cache permission errors, so I couldn’t complete runtime verification here.

PR #402, opened 2026-06-23
fix: hide dictation overlay from VoiceOver and fast-path system shortcuts (#401)

### LINKED ISSUE #401: VoiceOver: dictation overlay steals focus and Cmd+Tab becomes sluggish
## Summary

FluidVoice is unusable with VoiceOver enabled. Two distinct problems make it impossible to keep working in another app while dictating:

1. **Overlay panels announce to VoiceOver and can shift focus** — when dictation starts, VoiceOver announces the overlay window and the user loses their place in the working app.
2. **Cmd+Tab becomes sluggish** — app switching takes noticeably longer while FluidVoice is running, even when idle (not recording). This affects all app switching, not just switching to/from FluidVoice.

As a VoiceOver user, I deleted the app entirely before finding the repository and investigating.

## Status — PR #590

PR #590 implements the two fixes that live in the main app (the default **bottom** overlay path):

- ✅ **Root cause 1** (overlay panels visible to VoiceOver) — fixed in `BottomOverlayView.swift`.
- ✅ **Root cause 3** (CGEvent tap latency on system shortcuts) — fixed in `GlobalHotkeyManager.swift`. The fast-path was refined during review: it is scoped to Cmd+Tab / Cmd+Space (the ANSI-only backtick keycode was dropped because it differs on ISO layouts), it is skipped when the chord matches a configured Fluid shortcut so real bindings are never swallowed, and it preserves the modifier-only bookkeeping. The snippet under Root Cause 3 below is the original illustration, not the final implementation.
- ⏳ **Root cause 2** (DynamicNotchKit panel steals focus) — **not** in #590. That fix belongs in the separate `DynamicNotchKit` dependency repo and only affects the non-default **notch** overlay, so it is tracked separately. This issue stays open until the DynamicNotchKit fix is also handled.

## Environment

- macOS 26.5.1 (Sequoia)
- VoiceOver enabled
- FluidVoice 1.6.0 (build 12)
- Overlay position: Bottom (default)

### 1. Overlay panels visible to VoiceOver

The four floating `NSPanel` instances in `BottomOverlayView.swift` (main overlay, prompt selector, mode selector, actions selector) do not hide themselves from the accessibility tree. VoiceOver discovers and announces them, which can shift focus away from the user's working app.

**Fix:** Add `setAccessibilityElement(false)` and `setAccessibilityRole(.unknown)` to all panel creation sites, and add `.accessibilityHidden(true)` to the `BottomOverlayView` SwiftUI body.

### 2. DynamicNotchKit panel steals focus

`DynamicNotchPanel.swift` in the DynamicNotchKit dependency overrides `canBecomeKey` and `canBecomeMain` to return `true`. When the notch overlay is used, this allows the panel to become the key window and activate the app.

**Fix:** Change both overrides to return `false`. Also add `setAccessibilityElement(false)` and `setAccessibilityRole(.unknown)`.

### 3. CGEvent tap adds latency to system shortcuts

`GlobalHotkeyManager.swift` creates a CGEvent tap at `.headInsertEventTap` that intercepts **all** `keyDown`, `keyUp`, and `flagsChanged` events system-wide. The callback does significant work (modifier tracking, shortcut matching, spawning async tasks for `PostTranscriptionEditTracker`) before passing non-matching events through.

With VoiceOver also intercepting keyboard events, both handlers run on every keystroke. System shortcuts like Cmd+Tab pass through both the event tap callback and VoiceOver's processing, adding noticeable latency.

**Fix:** Add an early return at the top of `handleKeyEvent` that immediately passes through system shortcuts (Cmd+Tab, Cmd+Shift+Tab, Cmd+Space, Cmd+\`) before any processing:

```swift
// Fast path: pass through system shortcuts to avoid adding latency
// to app switching, especially with VoiceOver.
if type == .keyDown || type == .keyUp {
    let flags = event.flags
    if flags.contains(.maskCommand) {
        let kc = UInt16(event.getIntegerValueField(.keyboardEventKeycode))
        // Tab(48), Space(49), Backtick(50)
        if kc == 48 || kc == 49 || kc == 50 {
            return Unmanaged.passUnretained(event)
        }
    }
}
```

## Steps to Reproduce

1. Enable VoiceOver (Cmd+F5)
2. Open any app (e.g. TextEdit) and start typing
3. Press the FluidVoice dictation shortcut
4. **Problem 1:** VoiceOver announces the overlay, focus may shift
5. Stop dictation, then try Cmd+Tab between apps
6. **Problem 2:** Cmd+Tab feels sluggish/delayed compared to when FluidVoice is not running

## Expected Behavior

- Dictation overlay should be invisible to VoiceOver — no announcements, no focus changes
- Cmd+Tab and other system shortcuts should have zero added latency from FluidVoice's event tap

## Files to Change

- `Sources/Fluid/Views/BottomOverlayView.swift` — 4 panel creation sites + SwiftUI body
- `Sources/Fluid/Services/GlobalHotkeyManager.swift` — `handleKeyEvent` method
- DynamicNotchKit dependency: `Sources/DynamicNotchKit/Utility/DynamicNotchPanel.swift`

postoso:
Fixes #401.

Two FluidVoice-side fixes, working from @Alshekhi's patch in the issue.

## Overlay focus / VoiceOver
The recording overlay panels (main pill + the prompt/mode/actions submenus) and the overlay's SwiftUI content weren't hidden from the accessibility tree, so VoiceOver announced them and could shift focus away from the user's active app. Each panel now sets `setAccessibilityElement(false)` + `setAccessibilityRole(.unknown)`, and the SwiftUI body uses `.accessibilityHidden(true)`.

## Cmd+Tab / system-shortcut latency
The session event tap runs the full modifier-tracking + shortcut-matching pass on every key event system-wide, which adds perceptible latency to Cmd+Tab / Cmd+Space / Cmd+` (worse when VoiceOver is also intercepting). Added a fast path that passes these reserved system shortcuts straight through before any matching work. None of them can be a FluidVoice hotkey, so it's safe.

The fast path still cancels a pending modifier-only hold on key-down before passing through, so a Command-based modifier-only FluidVoice shortcut held during a Cmd+Tab isn't mistaken for a clean modifier hold.

## Out of scope
The issue's third root cause (`DynamicNotchPanel.canBecomeKey/canBecomeMain`) lives in the external `altic-dev/DynamicNotchKit` package, not this repo, so that's a separate change there.

## Testing
- xcodebuild build: succeeds
- 5 new unit tests covering the fast-path passthrough logic: pass
- swiftlint --strict: 0 violations

### DIFF
--- a/Fluid.xcodeproj/project.pbxproj
+++ b/Fluid.xcodeproj/project.pbxproj
@@ -13,6 +13,7 @@
 		7C3697892ED70F9C005874CE /* DynamicNotchKit in Frameworks */ = {isa = PBXBuildFile; productRef = 7C3697882ED70F9C005874CE /* DynamicNotchKit */; };
 		7C5AF14B2F15041600DE21B0 /* MediaRemoteAdapter in Frameworks */ = {isa = PBXBuildFile; productRef = 7C5AF14A2F15041600DE21B0 /* MediaRemoteAdapter */; };
 		7CDB0A2D2F3C4D5600FB7CAD /* DictationE2ETests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7CDB0A292F3C4D5600FB7CAD /* DictationE2ETests.swift */; };
+		7CFA0E022F3C4D5600FB7CAD /* GlobalHotkeyManagerSystemShortcutTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7CFA0E012F3C4D5600FB7CAD /* GlobalHotkeyManagerSystemShortcutTests.swift */; };
 		7CDB0A2E2F3C4D5600FB7CAD /* AudioFixtureLoader.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7CDB0A2A2F3C4D5600FB7CAD /* AudioFixtureLoader.swift */; };
 		7CDB0A2F2F3C4D5600FB7CAD /* dictation_fixture.wav in Resources */ = {isa = PBXBuildFile; fileRef = 7CDB0A2B2F3C4D5600FB7CAD /* dictation_fixture.wav */; };
 		7CDB0A302F3C4D5600FB7CAD /* XCTest.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 7CDB0A2C2F3C4D5600FB7CAD /* XCTest.framework */; };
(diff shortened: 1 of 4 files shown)


postoso:
Closing this one. To be clear, #401 is still open and unfixed, so this closing is not me saying the problem went away.

Two things moved underneath this branch since June:

1. The overlay now deliberately sets itself as an accessibility element (`Sources/Fluid/Views/BottomOverlayView.swift:141-142`, unset again at 505-506). That is the opposite of the `accessibilityHidden` approach in this PR, so replaying it would fight a decision already made upstream.

2. `GlobalHotkeyManager.swift` took 12 commits since this branched, including #815 rewriting release handling to bind to the exact active shortcut. That covers the same invariant my fast path was protecting, just differently.

Both halves would need rewriting rather than rebasing, so a fresh pass sized against current code makes more sense than reviving this.

If anyone picks it up: the VoiceOver half is worth doing on its own, and @Alshekhi's original patch on #401 is still the best starting point. The CGEvent tap latency work is a separate concern and probably not worth bundling unless you want it.

PR #403, opened 2026-06-23
fix(privacy): stop persisting dictated content to the disk log

postoso:
`DebugLogger` persists every log line to `~/Library/Logs/Fluid/Fluid.log` unconditionally: `FileLogger.append` runs before the `loggingEnabled` guard, so the "Enable debug logs" toggle only gates the in-memory UI panel, not the disk file. As a result, on release builds the dictated transcript, spoken prompts/instructions, selected text, and LLM request/response bodies were written to that log in plaintext. The dictation prompt-trace path was also on by default, so it logged the transcript + prompt + AI answer on every AI-enhanced dictation.

This redacts user content to metadata only (lengths, byte counts, model/provider names, status, response field names) at every affected site, keeping diagnostics useful. Full prompt traces are still available on the console behind `FLUID_PROMPT_TRACE=1`, but are never persisted to disk.

## Scope
28 logging sites across 9 files (`LLMClient`, `AIProvider`, `FunctionCallingProvider`, `ASRService`, the two Apple speech providers, `ContentView`, `RewriteModeService`, `TypingService`). Includes `LLMError.httpError`'s description, which embedded the raw provider error body and escaped to the log via downstream `error.localizedDescription` calls.

## Testing
- xcodebuild build: succeeds
- swiftlint --strict: 0 violations
- Completeness cross-checked (a remaining-leak re-grep + a cross-model review). No regression test added: the project has no SwiftPM test target and these surfaces are private, so a dedicated Xcode test target would be disproportionate; the completeness review is the gate.

### DIFF
--- a/Sources/Fluid/ContentView.swift
+++ b/Sources/Fluid/ContentView.swift
@@ -2147,7 +2147,7 @@ struct ContentView: View {
 
         // If this was a rewrite recording, process the rewrite instead of typing
         if wasRewriteMode {
-            DebugLogger.shared.info("Processing rewrite with instruction: \(transcribedText)", source: "ContentView")
+            DebugLogger.shared.info("Processing rewrite (instruction chars: \(transcribedText.count))", source: "ContentView")
             AnalyticsService.shared.recordModelUsage(
                 role: .transcription,
                 mode: .edit,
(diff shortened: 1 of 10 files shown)


postoso:
Closing this out. On reflection the scattered shape here isn't something a narrower version fixes: the content gets interpolated at ~20 call sites, so any correct fix touches at least as many files as this PR, and the version with a real regression test comes out larger. Rather than leave a stale diff adding to the queue, I'm withdrawing. Happy to revisit if logging privacy gets picked up on the roadmap.

PR #426, opened 2026-06-26
Add adjustable tail recording to capture the last word

EliseiNicolae:
## Description
Dictation was clipping the last word. When you trigger stop, the final ~50–130 ms of audio is still in CoreAudio's input pipeline (and people tend to release the key a hair early), but capture freezes immediately — so the tail of the last word is lost.

This keeps the mic tap live for a short, **configurable** grace period after the stop trigger, so the trailing audio lands in the buffer before the engine tears down. The duration is exposed as a slider in Settings.

## Type of Change
- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🧹 Chore
- [ ] 📝 Documentation update

## Related Issues
- N/A

## What changed
- **`ASRService.stop()`** now `await`s a short grace period *before* `audioCapturePipeline.setRecordingEnabled(false)`, so the tap keeps appending the trailing audio. The exact cessation point (the tap's `guard enabled` check) is why the delay has to go before that line — anything after it is already discarded.
- Guarded by a new **`isStopping`** flag to close the re-entrancy window the delay opens: while `stop()` is sleeping, `isRunning` is still `true`, so the un-guarded playground/Welcome stop path could otherwise slip in a second `stop()`.
- New persisted setting **`SettingsStore.recordingTailDuration`** (seconds, default **0.2**, clamped 0–0.4). Read live at stop time, so a change applies on the very next dictation. `0` disables the tail entirely (the getter uses a `nil`-check so `0` persists instead of snapping back to the default).
- Slider added under **Settings → Global Hotkey → Options → "Extra recording after stop"**, mirroring the existing "Bottom Offset" slider pattern.
- Wired through settings **backup/restore** as an optional field, so older backups still decode.
- The **Cancel** path (`stopWithoutTranscription()`) is intentionally left undelayed.

## Testing
- [ ] Tested on Intel Mac
- [x] Tested on macOS 26

Built Release (arm64) successfully, signed, installed, and launched on Apple Silicon / macOS 26. End-to-end dictation tail behavior (and dragging the slider) should be confirmed manually, since it needs live mic input.

## Notes
- **Latency tradeoff:** stop→text grows by the configured tail (default 0.2 s). Reasonable for catching the last word; the user can tune it down to 0 or up to 0.4 s.
- **Parakeet fast-preview interaction:** the extra tail audio more often trips the `tail_has_audio` guard, so finalize runs a full transcription instead of reusing the streaming preview — slightly slower finalize on that path, but that's exactly what recovers the last word.
- **No overlay flicker:** the overlay already flips to "Transcribing" before `stop()` is called, and `isProcessingActive` keeps it visible while `isRunning` flips a beat later.

## Screenshots / Video
The new control lives in **Settings → Global Hotkey card → Options**, directly under "Activation Mode" — label "Extra recording after stop", a 0.00–0.40 s slider showing the live value. _Screenshot to be added._

### DIFF
--- a/Sources/Fluid/Persistence/BackupService.swift
+++ b/Sources/Fluid/Persistence/BackupService.swift
@@ -81,6 +81,8 @@ struct SettingsBackupPayload: Codable, Equatable {
     let visualizerNoiseThreshold: Double
     let overlayPosition: SettingsStore.OverlayPosition
     let overlayBottomOffset: Double
+    // Optional so backups written before this field existed still decode.
+    let recordingTailDuration: Double?
     let overlaySize: SettingsStore.OverlaySize
     let transcriptionPreviewCharLimit: Int
     let userTypingWPM: Int
--- a/Sources/Fluid/Persistence/SettingsStore.swift
+++ b/Sources/Fluid/Persistence/SettingsStore.swift
@@ -2160,6 +2160,21 @@ final class SettingsStore: ObservableObject {
         }
     }
 
+    /// Extra seconds to keep recording after the user presses stop, so the tail of their
+    /// last word isn't clipped. Default 0.2; range 0–0.4s. Read by ASRService at stop time.
+    var recordingTailDuration: Double {
+        get {
+            // 0 is a valid value here (disables the tail), so don't treat a stored 0 as "unset".
+            guard self.defaults.object(forKey: Keys.recordingTailDuration) != nil else { return 0.2 }
+            return self.defaults.double(forKey: Keys.recordingTailDuration)
+        }
+        set {
+            objectWillChange.send()
+            let clamped = max(min(newValue, 0.4), 0.0)
+            self.defaults.set(clamped, forKey: Keys.recordingTailDuration)
+        }
+    }
+
     /// The size of the recording overlay (default: medium)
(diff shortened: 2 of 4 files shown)


EliseiNicolae:
Demo: 
https://drive.google.com/file/d/1KBdOcTNM3FDFH1E2nX-p5nRSeWpI3IR1/view?usp=sharing

altic-dev:
Amazing job making it configurable. Because I don't wanna ship it by default as people would mostly complain about how slow it is compared to other apps. So making it configurable helps deal with getting the last word I guess if needed. 

I wonder if we can look into optimizing coreAudio to capture it without the 100ms delay in the first place. 

Also, Nice demo. Shows it off immediately

altic-dev:
Okay, I did do some understanding of this issue and I feel like there are ways to optimize this better rather than patching this using an extra delay. I think it should capture everything automatically given you can add a 20 to 50 millisecond delay to capture everything as optional. So, this way, the user wouldn't have to worry about what to do or what timing to put, and things like that. Look into time-stamping of core audio frames and process based on that, it could be a better way to do this. If that doesn't really work or adds too much latency, then we can go forward with merging this

PR #429, opened 2026-06-26
fix(localapi): reject over-limit inline audio instead of silently truncating

postoso:
## Description
LocalAPI has a 300-second audio limit, but inline audio decoded through `LocalAPIAudioDecoder.samples(fromAudioData:suggestedExtension:)` could be silently truncated to the first 300 seconds because the shared decoder capped `framesToRead` with `min(file.length, maxFrames)`.

This changes the shared decode path to reject over-limit audio with the same `LocalAPIAudioDecoder` duration-limit error instead of truncating. Under-limit inline audio still decodes normally.

The PR also centralizes the duration-limit error in `durationLimitExceededError()`, so file-path and inline/body uploads return the same error domain, code, and message.

## Type of Change
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] 📝 Documentation update

Tracking discussion: #429.

## Testing
- [ ] Tested on Intel Mac
- [x] Tested on macOS 15.7.7
- [x] Ran linter locally: `swiftlint --strict --config .swiftlint.yml Sources Tests/FluidDictationIntegrationTests/LocalAPIAudioDecoderTests.swift`
- [x] Built locally: `xcodebuild build-for-testing -project Fluid.xcodeproj -scheme Fluid -destination 'platform=macOS,arch=arm64' CODE_SIGN_IDENTITY='' CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO`

Added `LocalAPIAudioDecoderTests` coverage:
- `testInlineAudioOverLimitThrowsInsteadOfTruncating`
- `testInlineAudioUnderLimitDecodesNormally`
- `testFilePathDecodeRejectsOverLimitAudio`

The tests build synthetic silent WAV fixtures/data, so they do not require downloading or running a speech model.

## Notes
This is API behavior only. It does not change UI, settings, onboarding, overlay, menu bar, or visual behavior.

Full hosted test execution was not run locally because the installed FluidVoice app was open; the hosted test runner can hang in that state. `build-for-testing` compiled the app and test target, and CI should run the full hosted suite after the branch update.

## Screenshots / Video

### DIFF
--- a/Fluid.xcodeproj/project.pbxproj
+++ b/Fluid.xcodeproj/project.pbxproj
@@ -13,6 +13,7 @@
 		7C5AF14C2F15041600DE21B0 /* MediaRemoteAdapter in Embed Frameworks */ = {isa = PBXBuildFile; productRef = 7C5AF14A2F15041600DE21B0 /* MediaRemoteAdapter */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		7C9A71022F58B00000FB7CAF /* TranscribeCpp in Frameworks */ = {isa = PBXBuildFile; productRef = 7C9A71012F58B00000FB7CAF /* TranscribeCpp */; };
 		7C91B0012F42AA0100C0DEF0 /* HotkeyShortcutTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7C91B0022F42AA0100C0DEF0 /* HotkeyShortcutTests.swift */; };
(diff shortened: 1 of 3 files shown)


postoso:
Closing this as superseded. Main no longer truncates audio at 300 seconds or enforces that limit.
`ChunkReader` reads the file in 20-minute chunks and the caller continues until EOF, so the
behaviour this PR fixed is gone and the rejection it added no longer matches main.

The tests here are tied to the removed duration limit, so there is nothing to rebase.

PR #431, opened 2026-06-26
fix(dictation): stop leaking raw <think> content when stripped output is empty

postoso:
## What

In the non-streaming LLM response parsers, `content` fell back to the RAW content when the thinking-stripped content was empty (`cleanedContent.isEmpty ? rawContent : cleanedContent`). When a reply is all thinking (`<think>...</think>` with nothing after), `cleanedContent` is empty, so the raw tagged text leaked downstream and got inserted into the user's document. The streaming path already returns empty (which trips the existing `emptyResponse` error), so the two paths disagreed.

## How

Return `cleanedContent` directly at both parser sites (`parseResponsesResponse`, `parseMessageResponse`). An all-thinking reply now yields empty content and trips the existing empty-response handling instead of leaking the tags, matching the streaming path. The two parsers are widened from `private` to internal so the new tests can reach them; behavior is otherwise unchanged.

## Tests

New `LLMClientThinkingTagTests` (5 methods): all-thinking yields empty, content after `</think>` is kept clean, plain content is unchanged, covering both the chat-completions and Responses API parsers. `xcodebuild test` is green and `swiftlint --strict` is clean.


grohith327:
Closing PR, please open a discussion.

PR #432, opened 2026-06-26
fix(command-mode): detect destructive commands in chained segments

postoso:
## What

`CommandModeService.isDestructiveCommand` applied its prefix checks to the whole command string, so a destructive command hidden after a shell chaining separator bypassed the confirm gate and auto-ran. Examples that returned false (auto-run): `echo ok && killall Finder`, `true; shred secret.txt`, `echo a || killall Finder`, `echo ok & killall Finder`, and a newline-chained `sudo reboot`.

## How

Split the command on chaining separators (`&&`, `||`, `;`, `&`, newline) into segments, trim each, and apply the existing detection per segment (factored into `isDestructiveSegment`). `&&` is normalized before the single `&` so it is not read as two background separators. Pipe (`|`) is intentionally not a split point: piped destructive commands stay matched by the existing `| rm` / `| sudo` / `| dd` substring patterns. Verified that `2>&1` and `sleep 10 & echo done` do not false-positive.

## Tests

New `CommandModeDestructiveCommandTests` (15 methods): each separator hiding a destructive command, safe chains, a backgrounded-safe command, and preserved single-command plus piped detection. `xcodebuild test` is green and `swiftlint --strict` is clean.

## Note

Focused follow-up to #428 (which hardens the same function for pipe-to-shell and output redirects). The two address independent gaps; this is structured so it rebases cleanly after #428 merges (the new per-segment `isDestructiveSegment` helper is where #428's added patterns belong).


postoso:
Closing this for now to keep the review queue smaller. This overlaps the same command-mode safety area as #434, and #434 is the cleaner branch to finish first. If the chained-command gap still needs a follow-up after #434 lands, I can reopen it as a fresh, focused PR on top of the merged command-mode changes.

PR #439, opened 2026-06-27
Fix: send AI enhancement instructions in system role, not user message

### LINKED ISSUE #388: [🐞 BUG] AI Request formed in a way that is much more likely to lead to failed instruction following (answering the transcript instead of cleaning it).

Kayaba-Attribution:
## Description

Fixes a bug where AI enhancement instructions were placed in the user message instead of the system prompt, causing strict role-separation models to answer the transcript rather than apply the instructions.

### Root cause — two independent sites

**1. `ContentView.processTextWithAI` (user-facing dictation path)**

The dictation branch hardcoded `systemPrompt = ""` and folded instructions + transcript into the user turn via `renderDictationUserMessage`. A comment in the code marked this as intentional ("dictation enhancement folds the prompt + transcript into a single user turn"), but it breaks any model that treats role separation strictly (e.g. Cerebras gpt-oss-120b, local models with system-prompt enforcement).

**2. `DictationPostProcessingService.process` (local API path — `/v1/postprocess`)**

Same pattern, same root cause. This path is called from `InferenceAPIController`, not from the main UI dictation flow.

### Fix

Both sites now consistently put the instruction text in the system role and the transcript alone in the user role:

```swift
// Before (dictation branch in ContentView)
systemPrompt = ""
userMessageContent = SettingsStore.renderDictationUserMessage(
    promptText: promptText,
    transcript: inputText
)

// After
let systemPrompt = promptText
let userMessageContent = inputText
```

The `isDictationCall` conditional in `ContentView` is removed — both branches now do the same thing. `DictationPostProcessingService` gets the same treatment.

The no-prompt case (prompt turned off) is unaffected: `effectiveDictationSystemPrompt` returns `""` when prompts are off, so the existing `if !systemPrompt.isEmpty` guard in the messages builder correctly omits the system entry.

## Type of Change

## Related Issues
- Closes #388

## Testing
- [x] Tested on macOS Sequoia 15
- [x] 4 unit tests added in `DictationSystemPromptTests`:
  - `effectiveDictationSystemPrompt` returns configured instruction text (not empty)
  - `effectiveDictationSystemPrompt` falls back to a non-empty built-in default
  - `renderDictationUserMessage("")` returns transcript only (no instructions bleeding in)
  - `${transcript}` placeholder substitution still works
- [ ] Tested on Intel Mac
- [x] Ran linter locally: `swiftlint --strict --config .swiftlint.yml Sources` — 0 violations
- [x] Ran formatter locally: `swiftformat --config .swiftformat Sources/Fluid/Services/DictationPostProcessingService.swift Sources/Fluid/ContentView.swift` — 0 additional changes

## Notes

Full end-to-end verification requires a live provider that exhibits the role-conflation failure (e.g. Cerebras gpt-oss-120b or a local model strict about system vs. user separation). The unit tests cover the message-composition logic without a network call.

### DIFF
--- a/Fluid.xcodeproj/project.pbxproj
+++ b/Fluid.xcodeproj/project.pbxproj
@@ -14,6 +14,7 @@
 		7C91B0012F42AA0100C0DEF0 /* HotkeyShortcutTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7C91B0022F42AA0100C0DEF0 /* HotkeyShortcutTests.swift */; };
 		7CDB0A2D2F3C4D5600FB7CAD /* DictationE2ETests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7CDB0A292F3C4D5600FB7CAD /* DictationE2ETests.swift */; };
 		7CDB0A2E2F3C4D5600FB7CAD /* AudioFixtureLoader.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7CDB0A2A2F3C4D5600FB7CAD /* AudioFixtureLoader.swift */; };
+		37C99EA57FCA4CDA8967073A /* DictationSystemPromptTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7C79B8A76E7C4F7A80A8EB95 /* DictationSystemPromptTests.swift */; };
 		7CDB0A2F2F3C4D5600FB7CAD /* dictation_fixture.wav in Resources */ = {isa = PBXBuildFile; fileRef = 7CDB0A2B2F3C4D5600FB7CAD /* dictation_fixture.wav */; };
 		7CDB0A302F3C4D5600FB7CAD /* XCTest.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 7CDB0A2C2F3C4D5600FB7CAD /* XCTest.framework */; };
 		7CE006BD2E80EBE600DDCCD6 /* AppUpdater in Frameworks */ = {isa = PBXBuildFile; productRef = 7CE006BC2E80EBE600DDCCD6 /* AppUpdater */; };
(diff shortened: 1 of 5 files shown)


altic-dev:
This is a tricky fix as it causes conflicts betweeen another PR about the state of system vs user prompt. Would to have a discussion before we merge this.

Kayaba-Attribution:
Okay, could you provide more task reqiuremens/concerns details? @altic-dev

Issue #467, opened 2026-06-29
Speech model silently resets to Whisper Tiny after an in-app update


postoso:
**What happens**

After updating FluidVoice (most recently to 1.6.1), my selected speech model was silently changed from Parakeet TDT v2 to Whisper Tiny. I didn't notice until transcriptions got bad (Whisper Tiny hallucinates text during silence, e.g. random music or tag-like tokens), then I found the model had been switched and set it back to Parakeet by hand.

I'm fairly sure this also happened on an earlier update (around 1.6.0, about a week ago), though I only have logs for the 1.6.1 occurrence. For a less technical user this is rough: they would just get bad transcriptions after an update with no idea why.

**Environment**
- FluidVoice 1.6.1 (build 12)
- macOS on Apple Silicon
- Updated via the in-app updater

**Evidence (today's 1.6.1 update)**

The app log right after the update shows dictations running on whisper-tiny, surviving an app relaunch (so the selection was persisted on disk, not just in memory), then my manual switch back to Parakeet TDT v2 a few minutes later. My persisted SelectedSpeechModel reads parakeet-tdt-v2 again now.

**What I ruled out**

I went through the source to save you time. None of these is the cause:
- The 1.6.1 model-cache cleanup never writes SelectedSpeechModel, and my Parakeet cache was intact.
- Legacy migration: I have no legacy whisper defaults keys (SelectedTranscriptionProvider / WhisperModelSize), so migration could only return the Parakeet default here, never whisper-tiny.
- Onboarding: whisper-tiny is not an onboarding route candidate for any language (the whisper option there is whisper-small), so even the forced re-onboarding on a generation bump cannot produce it.
- The getter's platform fallback returns whisper-base, not tiny, and is read-only (never persisted).

As far as I can tell, whisper-tiny is unreachable through any automatic path. The only code that writes it is a direct selection: tapping Activate on the Whisper Tiny card, the deleteSpeechModel temp-swap, or a backup restore. None of those run on their own at launch.

**Best guess**

The only update-specific code is the updater relaunch, which briefly runs two app instances (createsNewApplicationInstance = true, with the old instance terminated about 2 seconds later). During that window two processes share the same UserDefaults domain and the SettingsStore singleton, so a multi-process write could be involved. I could not prove it lands on whisper-tiny specifically, so treat this as a hypothesis, not a conclusion.

**The blocker, and an offer**

[...]

PR #468, opened 2026-06-29
fix: do not persist a deleted speech model as the active selection

postoso:
## Description
Deleting a non-active speech model used to briefly persist the model being deleted as `selectedSpeechModel`, then restore the real selection through an un-awaited async restore. If the app quit, crashed, or relaunched during that window, the deleted model could become the saved active model.

This routes per-model deletion through an awaitable `ASRService.clearModelCache(for:)` call, which clears the target model without changing the saved selection. It also refreshes installed-model state after deletion so the model cards update without relying on a temporary selection change.

The removed `suppressSpeechProviderSync` and `skipNextSpeechModelSync` flags were only needed for the old selection bounce. With the bounce removed, the speech provider and preview state can stay directly synced to the actual selected model. The view-model method is now awaitable, so the regression test does not rely on a fixed sleep.

## Type of Change
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] 📝 Documentation update

Related to #467. This fixes the delete-non-active-model persistence path only; it does not claim to close every possible path that could reset the active speech model.

## Testing
- [ ] Tested on Intel Mac
- [x] Tested on macOS 15.7.7
- [x] Ran linter locally: `swiftlint --strict --config .swiftlint.yml Sources Tests/FluidDictationIntegrationTests/DictationE2ETests.swift`
- [x] Built locally: `xcodebuild build-for-testing -project Fluid.xcodeproj -scheme Fluid -destination 'platform=macOS,arch=arm64' CODE_SIGN_IDENTITY='' CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO`

Added `DictationE2ETests.testDeleteNonActiveSpeechModel_doesNotBounceSelectionThroughDeletedModel`. The test records every write to `SelectedSpeechModel` and fails if deleting `.appleSpeech` while `.whisperTiny` is active ever persists the deleted value, even transiently.

The test intentionally uses a built-in no-cache model so it can catch the selection-bounce regression without deleting a real downloaded speech model from the developer's machine.

## Notes
This is intentionally narrow. It removes one concrete race in model deletion without changing model activation, update behavior, or broader speech-model fallback logic.

Full hosted test execution was not run locally because the installed FluidVoice app was open; the hosted test runner can hang in that state. `build-for-testing` compiled the app and test target, and CI should run the full hosted suite after the branch update.

## Screenshots / Video

### DIFF
--- a/Sources/Fluid/UI/AISettings/VoiceEngineSettingsViewModel.swift
+++ b/Sources/Fluid/UI/AISettings/VoiceEngineSettingsViewModel.swift
@@ -28,8 +28,6 @@ final class VoiceEngineSettingsViewModel: ObservableObject {
     @Published var selectedSpeechProvider: SettingsStore.SpeechModel.Provider
     @Published var previewSpeechModel: SettingsStore.SpeechModel
     @Published var showAdvancedSpeechInfo: Bool = false
-    @Published var suppressSpeechProviderSync: Bool = false
-    @Published var skipNextSpeechModelSync: Bool = false
 
     var downloadingModel: SettingsStore.SpeechModel? {
         guard let modelID = self.asr.downloadingModelId else { return nil }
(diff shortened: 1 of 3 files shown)


postoso:
Closing this one. It has drifted into a conflict with main, and the case that prompted it turned out to come from my own test setup rather than normal use, so I don't think it's worth your review time. The branch stays on my fork if it ever comes up again.

PR #489, opened 2026-06-30
Add opt-in 'Lock input device' setting

funkju:
## Summary

FluidVoice always followed the macOS system input device — Settings even states *"Audio devices are synced with macOS System Settings."* There was no way to pin dictation to a specific microphone, so any system input change (e.g. a Bluetooth headset connecting) would silently switch the mic used for capture.

The root cause is that `SettingsStore.syncAudioDevicesWithSystem` is hard-coded to `return true`, permanently disabling the existing preferred-input binding path in `ASRService`.

This PR adds an opt-in **"Always use this input device when available"** toggle (default **off**, so existing behavior is unchanged). When enabled, FluidVoice binds the chosen input device on the audio engine and ignores macOS input changes. If the locked device is unavailable, it transparently falls back to the system default and re-binds to the locked device when it reconnects.

## Changes

- **`SettingsStore`** — new opt-in `lockInputDevice` setting (`LockInputDevice` key).
- **`ASRService`** — `bindPreferredInputDeviceIfNeeded()` now also binds when `lockInputDevice` is on; `handleDefaultInputChanged()` and `handleDeviceListChanged()` skip the auto-switch/route-recovery logic while locked.
- **`SettingsView`** — new toggle (far-right, switch style) with concise help text; input picker no longer overrides the system default while locked; restores the locked device on appear. Refreshes the *(System Default)* marker via the audio observer's `changeTick` so it stays accurate while locked.
- **`ContentView`** / **`MenuBarManager`** — keep the preferred input selected (instead of the system input) while locked.

## Design notes

- **Input-only by intent.** This deliberately does not touch output-device locking, which hits `OSStatus -10851` (`kAudioUnitErr_InvalidPropertyValue`) for aggregate/Bluetooth devices.
- **Default off.** No behavior change unless the user opts in, matching the existing `syncAudioDevicesWithSystem` contract; nothing new is added to backup/restore for the same reason.

## Testing

Built locally (ad-hoc signed) and verified: locking pins capture to the selected mic, system input changes are ignored while locked, and disconnect/reconnect of the locked device falls back and recovers correctly.

### DIFF
--- a/Sources/Fluid/ContentView.swift
+++ b/Sources/Fluid/ContentView.swift
@@ -347,8 +347,15 @@ struct ContentView: View {
                 // Only sync UI with system defaults when sync is enabled
                 // When sync is disabled, keep the user's preferred device selection
                 if SettingsStore.shared.syncAudioDevicesWithSystem {
-                    // Sync mode: Update UI to match current system defaults
-                    if let sysIn = AudioDevice.getDefaultInputDevice()?.uid {
+                    // Sync mode: Update UI to match current system defaults.
+                    // Exception: when the input device is locked, keep the pinned input selection
+                    // (falling back to the system default only if the pinned device disappeared).
+                    if SettingsStore.shared.lockInputDevice,
+                       let prefIn = SettingsStore.shared.preferredInputDeviceUID,
+                       inputDevices.contains(where: { $0.uid == prefIn })
+                    {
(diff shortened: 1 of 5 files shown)


grohith327:
Hey @funkju thanks for the PR, resolving this in favor of #531 since it adds a few more functionality and tested with the latest transcription pipeline

Issue #380, opened 2026-06-12
[✨ FEATURE] Use specific phrase like "press enter" to send the enter keystroke


stathius:
**Is your feature request related to a problem?**
I want to vibe code without touching my keyboard.

**Describe the solution you'd like**
se specific phrase like "press enter" to send the enter keystroke.

**Describe alternatives you've considered**
Press enter manually.

**Additional context**
NA

altic-dev:
new model coming up that works with 'new line' and 'new paragraph' actually!

if you want smth for now, you can add 'press enter' to your custom dictationary -> instant replacement  and possibly find the character for new line and add it there! Thanks.

altic-dev:
printf '\n' | pbcopy

copy this into your terminal - it copies what a new line is supposed to be to your clipboard 

now go to instant replacement in settings and custom dictionary. 

<img width="494" height="398" alt="Image" src="https://github.com/user-attachments/assets/8b3c5646-ef8b-48c2-931f-28e92e3d2f3e" />


go here and paste it in replacement. [ It will look empty but give it a try in your dictation, it will create new lines ]

PR #524, opened 2026-07-05
Fix login launch (#520), volume spike (#522), add spoken Enter keystr…

### LINKED ISSUE #520: [BUG] Persistent & invisible Background Launch on System Startup
### Describe the bug

Currently, when the app is configured to launch on system startup (or added manually via macOS System Settings), it does not launch seamlessly. Upon restarting the Mac, the application often fails to start entirely, or when it does launch, it opens the main application window repeatedly/unnecessarily.

This forces a manual intervention after every reboot to ensure the app is running and to close the unwanted window, which creates a significant friction point in the user experience.

### Reproduction steps

1. Restart computer and see that FluidVoice show the window of app or not run

### Expected behavior

The application should launch silently in the background upon system startup. It should reside in the system status bar without displaying any GUI windows unless explicitly invoked by the user.

### Actual behavior

As described above

### App Version

1.6.1

### macOS Version

26.5.1 

### Architecture

Apple Silicon

### Logs or crash report

```text

```

_No response_

### LINKED ISSUE #522: Independent volume selection causes temporary system volume spike
## Description

When making an independent volume selection, there is a temporary spike in system volume that can be startling — especially if youre not expecting it.

## Steps to Reproduce

1. Make an independent volume selection
2. Observe the system volume behavior

## Expected Behavior

Volume changes should be smooth and predictable, without any sudden spikes.

## Actual Behavior

A brief but noticeable spike in system volume occurs during the selection, which can cause an unpleasant surprise.

## Environment

- **Device:** MacBook Pro M5
- **OS:** macOS 27
- Recorded this ticket using Fluid Voice — it rocks! 🎙️

## Additional Context

I saw the mention of it impacting system volume, but I wasnt expecting it to be that intense. The spike is brief but loud enough to catch you off guard. Would be great to either smooth out the transition or cap the peak during selection.

### LINKED ISSUE #523: [✨ FEATURE] Native Enter keypress emulation via spoken phrase (works in terminals)
### Problem / motivation

The maintainer's workaround in #380 is to copy a real newline (\`printf '\n' | pbcopy\`) and map a Custom Dictionary entry \`press enter\` → paste the newline into *Correct Spelling*. This works in chat apps, but has two limits:

1. **The replacement field won't retain a newline.** After pasting the copied \`\n\` into *Correct Spelling (replacement)* and saving, the field stays empty and nothing is inserted on dictation — the trigger is just deleted. The input appears to strip whitespace-only values, so the workaround can't actually be saved.

2. **It's a paste, not a keystroke, so terminals ignore it.** Instant Replacement inserts text via clipboard paste. Modern shells (zsh/bash on macOS) enable **bracketed paste mode** by default, which treats pasted newlines as literal characters and intentionally does *not* execute the command. So "say *press enter* to submit" is a no-op in Terminal/iTerm.

### Request

A native, system-level **Enter keypress emulation** — a real \`Return\` keystroke injection triggered by a spoken phrase — instead of inserting a \`\n\` character. This would submit uniformly across chat apps **and** terminals, and wouldn't depend on the (broken) newline-in-dictionary workaround.

Overlaps with #518 (auto-submit after paste) but from the spoken-command angle of #380.

### Environment
macOS, zsh (bracketed paste on by default).

TheJagStudio:
## Description
Addresses three issues in one focused change set (6 files):

**#523 – Native Enter keypress emulation via spoken phrase.** New opt-in setting "Press Enter by Voice" with a configurable trigger phrase (default `press enter`; falls back to the default if left empty, so the empty-replacement problem from the dictionary workaround in #380 can't occur). When the transcript contains the phrase, `TypingService` splits the output into segments and posts a real `kVK_Return` CGEvent keystroke (targeted at the focused app's PID, HID tap fallback) where the phrase was spoken, instead of pasting a `\n`. Because it is a genuine keystroke, zsh/bash with bracketed paste execute the command — this works in terminals as well as chat apps. Matching is case-insensitive, swallows punctuation the transcriber attaches around the phrase, handles a transcript consisting only of the phrase, and waits ~300 ms after paste-based insertion so the Return never fires before the text lands. The new settings round-trip through backup export/import.

**#522 – Independent volume selection causes system volume spike.** `TranscriptionSoundPlayer` previously always set the system volume to the cue's level and played at full player volume, spiking any other audio. Now, when the desired cue level fits under the current system volume (the common case), the system volume is never touched — loudness is compensated inside the player (`playerVolume = desired / systemVolume`). Only when the cue is louder than the system volume does it boost, and the boost now ramps in 8 small steps over ~120 ms up and back down instead of jumping. Overlapping sounds no longer clobber the saved pre-boost volume, so restores can't get stuck at a boosted level. Settings footnote updated accordingly.

**#520 – Persistent & invisible background launch on system startup.** Login-launch detection relied on the legacy `keyAELaunchedAsLogInItem` Apple Event, which SMAppService login launches on modern macOS often don't carry — so every reboot looked like a manual launch and the main window appeared. Detection now falls back to a heuristic: the app is registered as a login item and the process started within 5 minutes of boot. The default for "Show window when launched at login" is flipped to OFF so login launches start silently in the menu bar (manual launches always show the window), and the LaunchServices reopen fallback now boots the window hidden immediately instead of flashing it until the retry loop catches it.

## Type of Change
- [x] 🐞 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 🧹 Chore
- [ ] 📝 Documentation update

Closes #520
Closes #522
Closes #523

## Testing
- [ ] Tested on Intel Mac
- [ ] Ran tests locally:

## Screenshots / Video
This PR does change Settings UI: two new rows ("Press Enter by Voice" toggle + "Trigger phrase" field) and an updated footnote on "Independent Volume".

image

Note: `SettingsStore.swift` is flagged by the policy check but contains only persistence changes (new UserDefaults keys, backup fields, a default-value flip) — no visual code. The visual changes are the `SettingsView.swift` rows shown above.

- [ ] No UI/visual changes; screenshots/video are not applicable.

## Notes
- The `showMainWindowAtLoginLaunch` default flip means existing installs that never touched the toggle will now start silently at login (per #520's expected behavior); users can re-enable the window in App Settings.
- The login-launch heuristic (SMAppService enabled + uptime < 300 s) can classify a manual launch made within 5 minutes of boot as a login launch; the trade-off favors never flashing the window on reboot.
- "Press Enter by Voice" is off by default, so no behavior changes unless a user opts in.

### DIFF
--- a/Sources/Fluid/AppDelegate.swift
+++ b/Sources/Fluid/AppDelegate.swift
@@ -8,6 +8,7 @@
 import AppKit
 import Carbon
 import PromiseKit
+import ServiceManagement
 import SwiftUI
 import UserNotifications
 
@@ -94,6 +95,12 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
     func applicationShouldHandleReopen(_ sender: NSApplication, hasVisibleWindows flag: Bool) -> Bool {
         if self.shouldSuppressNextReopenActivation {
             self.shouldSuppressNextReopenActivation = false
+            // This reopen came from our own silent login-launch fallback. The reopen
+            // realizes the SwiftUI main window, so immediately boot it hidden instead of
+            // letting it flash on screen until the launch retry loop catches it (#520).
+            DispatchQueue.main.async { [weak self] in
+                _ = self?.bootMainWindowHiddenIfPresent()
+            }
             return true
         }
 
@@ -135,10 +142,31 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
         if ProcessInfo.processInfo.environment["FLUID_SIMULATE_LOGIN_LAUNCH"] == "1" {
             return true
         }
-        guard let event = NSAppleEventManager.shared().currentAppleEvent else { return false }
-        return event.eventID == AEEventID(kAEOpenApplication)
-            && event.paramDescriptor(forKeyword: AEKeyword(keyAEPropData))?.enumCodeValue
-            == OSType(keyAELaunchedAsLogInItem)
+
+        if let event = NSAppleEventManager.shared().currentAppleEvent,
+           event.eventID == AEEventID(kAEOpenApplication),
+           event.paramDescriptor(forKeyword: AEKeyword(keyAEPropData))?.enumCodeValue
+           == OSType(keyAELaunchedAsLogInItem)
+        {
+            return true
+        }
+
+        // SMAppService login launches on modern macOS frequently arrive WITHOUT the
+        // legacy login-item Apple Event flag, which made the app treat login launches
+        // as manual ones and show the main window on every reboot (issue #520).
+        // Heuristic fallback: FluidVoice is registered as a login item and the process
+        // started shortly after boot - a manual launch that early is very unlikely.
+        if SMAppService.mainApp.status == .enabled,
+           ProcessInfo.processInfo.systemUptime < 300
+        {
+            DebugLogger.shared.info(
+                "Login-item launch inferred (SMAppService enabled, uptime \(Int(ProcessInfo.processInfo.systemUptime))s)",
+                source: "AppDelegate"
+            )
+            return true
+        }
+
+        return false
     }
 
     /// Apply the user's dock-visibility preference ("Hide from dock", issue #162).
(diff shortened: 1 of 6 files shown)


grohith327:
Hey @TheJagStudio There are a three issues tackled in this PR. I would recommend following up in the issues and tackling them one at a time

Issue #479, opened 2026-06-29
Dictated list not inserted in Claude Code under cmux — only "a" appears (transcript correct; VS Code works)


saimageshvar:
### Description
When I dictate a **list**, FluidVoice normally formats it as an ordered/unordered list,
inserting the items one by one. This works in **VS Code** and other apps.

In **Claude Code running inside cmux**, this list insertion fails — instead of the list, only
the single character `a` gets inserted. The dictation itself is correct: the **Transcription
History** shows the full, correctly-generated list.

Non-list dictation is not the focus here — this issue is specifically about **dictating lists**
into Claude Code under cmux.

### Steps to Reproduce
1. Launch Claude Code inside **cmux** (a terminal emulator / tmux-compatible environment).
2. Focus the Claude Code input.
3. Trigger dictation in FluidVoice and speak a **list** (e.g. "first item, second item, third item").
4. Observe what gets inserted into Claude Code.
5. Open FluidVoice → Transcription History and compare.

### Expected Behavior
The dictated list is inserted as an ordered/unordered list (items one by one), matching the
Transcription History — the same way it works in VS Code.

### Actual Behavior
Only the character `a` is inserted instead of the list. The Transcription History contains the
full, correct list — so transcription succeeds but list insertion into the cmux terminal fails.

### Works / Doesn't Work
- ✅ VS Code — dictated list inserted correctly (formatted, item by item)
- ✅ Other apps — work as expected
- ❌ Claude Code running in cmux — only `a` is inserted instead of the list

### Environment
- FluidVoice: **1.6.1 (build 12)**
- macOS: **26.5.1 (25F80)**, Apple Silicon (arm64)
- Target app: Claude Code running inside **cmux**

### Additional Notes
Since the transcript is correct in history but list insertion is truncated to a single
character, this looks like a **list-formatting / text-insertion issue** specific to the cmux
terminal emulator. cmux uses a tmux-compatible terminal layer, and the multi-line / item-by-item
insertion FluidVoice uses for lists (simulated keystrokes including newlines, or clipboard
paste) may be getting consumed/truncated differently than in a native text field or VS Code's
input. It may be worth testing how FluidVoice's list-insertion path behaves against terminal
emulators specifically.

altic-dev:
Sometimes the paste fails in tmux related apps and docs. Please use clipboard based method for now :) it's in preferences and paste method. 

[...]

PR #526, opened 2026-07-05
fix: exclude temporary paste writes from clipboard manager history

postoso:
## Description

FluidVoice temporarily writes dictated text to `NSPasteboard.general` when using Reliable Paste or a paste fallback. Clipboard managers can observe that write before FluidVoice restores the previous clipboard, leaving an unwanted transcription in history.

This marks the shared temporary pasteboard item with `org.nspasteboard.TransientType` and `org.nspasteboard.AutoGeneratedType`. The normal string remains available to Cmd+V, and the existing clipboard restoration behavior is unchanged. `ConcealedType` is intentionally omitted because the transcription is not necessarily sensitive content.

The Reliable Paste setting description is also updated so it no longer warns about the clipboard-history behavior fixed here.

## Type of Change

- [x] Bug fix (non-breaking change)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

Related to #479, which relies on the clipboard paste path this change improves.

## Testing

- [x] Apple Silicon Mac on macOS 15.7.7
- [x] SwiftLint strict: 0 violations in the changed Swift files
- [x] `xcodebuild build-for-testing`
- [x] 4 focused `TypingServiceTransientPasteboardTests`
- [x] Verified the markers survive a real `NSPasteboard` write
- [x] End-to-end Raycast check: FluidVoice inserted the transcription into TextEdit and Raycast returned no clipboard-history result for that text

## Screenshots / Video

End-to-end verification: https://github.com/user-attachments/assets/bdbecb8f-ca65-4d07-8086-591583f22069

There are no UI layout changes.


altic-dev:
this seems like a good one! wanna fix all of the issues and perhaps add a clip for me to verify ? I can happily merge this ;) might be useful for me as well.

postoso:
Rebased this onto current `main` and resolved the old conflicts.

The temporary pasteboard write now carries the normal string plus the Transient and AutoGenerated markers. Concealed is intentionally omitted. I also updated the stale Reliable Paste description and added coverage confirming the markers survive a real `NSPasteboard` write.

Verified with SwiftLint, `build-for-testing`, 4 focused tests, and an end-to-end Raycast check. FluidVoice inserted the transcription into TextEdit, restored the clipboard, and Raycast returned no clipboard-history result for the inserted text.

Video attached below.


https://github.com/user-attachments/assets/bdbecb8f-ca65-4d07-8086-591583f22069

PR #543, opened 2026-07-07
fix(audio): break permanent main-thread deadlock on default device change while idle

### LINKED ISSUE #542: [BUG] Permanent main-thread deadlock on default audio device change while idle (AVAudioEngine dealloc vs AVAudioEngineConfigurationChange observer)
### Describe the bug

FluidVoice permanently freezes (spinning beachball on its windows, menu bar item unresponsive, 0% CPU) due to a main-thread deadlock triggered by a default audio device change while the app is idle. The app never recovers; force quit is the only way out.

I captured a `sample` of the hung process, and every one of the 2,557 snapshots shows the same two threads waiting on each other:

**Thread A — main thread.** The CoreAudio default-device-change listener (dispatched to main from `registerDefaultDeviceChangeListener()`) runs `scheduleAudioRouteRecovery` → `retireAudioEngine(reason: "idle_route_change:...")`. Inside `retireAudioEngine`, the final strong reference to the old `AVAudioEngine` is released on the main thread. `-[AVAudioEngine dealloc]` then does a synchronous dispatch onto the engine's internal `engine` serial queue and blocks in `__DISPATCH_WAIT_FOR_QUEUE__`.

**Thread B — the engine's internal `engine` queue.** The same device change is being handled by `AVAudioEngineImpl::IOUnitConfigurationChanged()`, which posts `AVAudioEngineConfigurationChange` via `NSNotificationCenter`. Because ASRService registers that observer with `queue: .main` ([ASRService.swift#L2317-L2322](https://github.com/altic-dev/FluidVoice/blob/v1.6.2/Sources/Fluid/Services/ASRService.swift#L2317-L2322)), notification delivery wraps the handler in an `NSOperation` on the main queue and blocks the posting thread in `-[NSOperation waitUntilFinished]` until the main thread runs it.

Thread A is waiting for Thread B's queue; Thread B is waiting for Thread A's queue. Circular wait, so the hang is permanent.

Two observations from the source that may help:

1. `retireAudioEngine` already tries to avoid exactly this by deferring the release of `engineStorage` to a background queue ([ASRService.swift#L605-L610](https://github.com/altic-dev/FluidVoice/blob/v1.6.2/Sources/Fluid/Services/ASRService.swift#L605-L610)). The sample shows `-[AVAudioEngine dealloc]` occurring on the main thread directly under `retireAudioEngine` regardless, so some other strong reference to the engine is being dropped synchronously inside that function (Swift release/inlined frames don't show in the sample, so I can't say which one from the trace alone).
2. The `AVAudioEngineConfigurationChange` observer body only spawns a `Task { @MainActor ... }`, so registering it with `queue: nil` (synchronous delivery on the posting thread) looks safe and would remove the `waitUntilFinished` on the main queue — breaking the cycle from the other side even if a main-thread dealloc slips through again.

### Reproduction steps

It is a race between the app's route-recovery teardown and the engine's own configuration-change handling, so it does not reproduce on demand. Conditions when it hit:

1. FluidVoice running and idle (not recording), "sync audio devices with system" enabled.
2. The system default input/output device changes (in my case most likely a Bluetooth headset connecting or disconnecting).
3. If the engine's internal queue is inside `IOUnitConfigurationChanged()` posting the configuration-change notification at the moment the main thread releases the engine in `retireAudioEngine`, the app deadlocks.

### Expected behavior

The idle route change retires and re-prewarms the audio engine; the app keeps working.

### Actual behavior

The app freezes permanently: UI unresponsive, 0% CPU, all samples identical over a 3-second `sample` window. It stayed in this state indefinitely (hours) until I killed the process. Plain SIGTERM was enough to terminate it; after relaunch everything worked normally.

### App Version

1.6.2 (build 13)

### macOS Version

macOS 26.5.1 (25F80)

### Architecture

Apple Silicon (arm64)

### Logs or crash report

Main thread (Thread A), trimmed from `sample` output — identical in all 2,557 snapshots:

```
2557 Thread_2049822   DispatchQueue_1: com.apple.main-thread  (serial)
  ...
  2557 _dispatch_call_block_and_release  (in libdispatch.dylib)
    2557 <deduplicated_symbol>  (in FluidVoice)
      2557 partial apply for closure #1 in closure #1 in ASRService.registerDefaultDeviceChangeListener()  (in FluidVoice)
        2557 ASRService.scheduleAudioRouteRecovery(reason:)  (in FluidVoice)
          2557 ASRService.retireAudioEngine(reason:)  (in FluidVoice)
            2557 -[AVAudioEngine dealloc]  (in AVFAudio)
              2557 _dispatch_sync_f_slow  (in libdispatch.dylib)
                2557 __DISPATCH_WAIT_FOR_QUEUE__  (in libdispatch.dylib)
                  2557 _dispatch_event_loop_wait_for_ownership  (in libdispatch.dylib)
                    2557 _dispatch_kq_poll  (in libdispatch.dylib)
                      2557 kevent_id  (in libsystem_kernel.dylib)
```

Engine's internal queue (Thread B):

```
2557 Thread_2967515   DispatchQueue_23295: engine  (serial)
  ...
  2557 _dispatch_block_async_invoke2  (in libdispatch.dylib)
    2557 invocation function for block in AVAudioEngineImpl::IOUnitConfigurationChanged()  (in AVFAudio)
      2557 -[NSNotificationCenter postNotificationName:object:userInfo:]  (in Foundation)
        2557 _CFXNotificationPost  (in CoreFoundation)
          2557 -[NSOperation waitUntilFinished]  (in Foundation)
            2557 _pthread_cond_wait  (in libsystem_pthread.dylib)
              2557 __psynch_cvwait  (in libsystem_kernel.dylib)
```

All other threads were idle (run loops in `mach_msg`, workers in `semaphore_wait`/`workq_kernreturn`). I have the full `sample` output and can attach it if useful.

devzahirul:
## Description
Fixes the permanent main-thread deadlock reported in #542: a system default audio device change while FluidVoice is idle (with "sync audio devices with system" enabled) could beachball the app forever at 0% CPU, recoverable only by force quit.

The `sample` attached to the issue shows a circular wait between two threads, and both legs live in `ASRService`:

1. **Main thread** — the CoreAudio default-device listener runs `scheduleAudioRouteRecovery` → `retireAudioEngine(reason: "idle_route_change:...")`, and the *final* strong reference to the old `AVAudioEngine` gets released on main. `-[AVAudioEngine dealloc]` then does a synchronous hop onto the engine's internal serial queue and blocks.
2. **Engine's internal queue** — the same device change makes the (idle-but-prewarmed) engine run `IOUnitConfigurationChanged()`, which posts `AVAudioEngineConfigurationChange`. Because our observer was registered with `queue: .main`, NotificationCenter wraps delivery in an `NSOperation` and blocks the *posting* thread in `waitUntilFinished` until the main thread runs it — which it never can.

This PR fixes both legs independently:

- **Observer (`registerEngineConfigurationChangeObserver`)**: register with `queue: nil` so delivery is synchronous on the posting thread and the engine's queue never waits on main. The handler body only spawns a `Task { @MainActor ... }`, which is safe from any thread, so behavior is unchanged. This eliminates the whole deadlock class — any main-thread call that synchronizes with the engine queue (`dealloc`, `stop()`, `removeTap`) can no longer wait forever on a blocked post.
- **Retirement (`retireAudioEngine`)**: the existing hand-off (`DispatchQueue.global(qos: .utility).async { _ = oldEngine }`) answers the question left open in the issue ("some other strong reference is being dropped synchronously"). There is no other reference — the hand-off itself is racy. Dispatching the block only guarantees the utility queue holds *a* reference, not the *last* one: when the trivial block finishes before `retireAudioEngine` returns (likely, since main still runs the DebugLogger call), the caller's own local performs the final release and dealloc lands back on main, directly under the `retireAudioEngine` frame — exactly what all 2,557 sample snapshots show. The fix moves the retired engine into a small `RetiredAudioEngineReference` holder so no main-thread local ever owns it past the dispatch; every main-side release happens-before the enqueue, so the drain on the utility queue is provably the final release.

## Type of Change
- [x] 🐞 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 🧹 Chore
- [ ] 📝 Documentation update

Fixes #542

## Testing
- [ ] Tested on Intel Mac
- [x] Tested on macOS version: 26.1
- [x] Ran linter locally: `swiftlint --strict --config .swiftlint.yml` (0 violations)
- [x] Ran tests locally: full `xcodebuild test` suite passes on `platform=macOS,arch=arm64` (all suites green)

The deadlock itself is a scheduling race between engine retirement and the engine's own configuration-change handling, so it cannot be reproduced on demand (the reporter hit it via Bluetooth headset connect/disconnect). The fix was validated by matching both changed code paths against the two blocked stacks in the issue's `sample` output, plus a clean build and full test run.

## Screenshots / Video

## Notes
- `queue: nil` is intentionally load-bearing and now documented inline: block-based observers on an explicit queue make `postNotification` wait (`-[NSOperation waitUntilFinished]` in the issue's sample), so the engine's internal queue must never target main. Semantics are preserved because the handler already hopped to the main actor via `Task`; delivery order relative to the recovery guards (`isRunning` / `isRecoveringAudioRoute`) is unaffected.
- `RetiredAudioEngineReference` follows the existing pattern in this file for cross-thread helpers (`private final nonisolated class ... : @unchecked Sendable`, cf. `AudioCapturePipeline`). It is created on main and touched exactly once by the draining block; the dispatch enqueue provides the ordering.
- Considered and rejected: keeping `queue: .main` and filtering by `object:` (the notification legitimately comes from our own engine, so the wait would remain), and moving `engine.stop()` off main (unnecessary once the posting thread can no longer block on main, and it would change engine state timing during recovery).
- No regression test is included: forcing `IOUnitConfigurationChanged()` to post at the exact moment of retirement requires real device-route changes and cannot be scripted deterministically in CI without flakiness.

### DIFF
--- a/Sources/Fluid/Services/ASRService.swift
+++ b/Sources/Fluid/Services/ASRService.swift
@@ -603,10 +603,17 @@ final class ASRService: ObservableObject {
         }
         self.audioCapturePipeline.clearPreroll()
 
-        let oldEngine = self.engineStorage
-        self.engineStorage = nil
-        if let oldEngine {
-            DispatchQueue.global(qos: .utility).async { _ = oldEngine }
+        // The final strong reference must be released off the main thread:
+        // -[AVAudioEngine dealloc] waits on the engine's internal serial queue,
+        // which can deadlock main against a concurrent configuration-change post
+        // (#542). Capturing a local in the async block is not enough — if the
+        // block finishes before this function returns, the local's release
+        // becomes the final one and dealloc lands back on main. The holder keeps
+        // the engine out of main-thread locals entirely.
+        if self.engineStorage != nil {
+            let retired = RetiredAudioEngineReference(self.engineStorage)
+            self.engineStorage = nil
+            DispatchQueue.global(qos: .utility).async { retired.drain() }
         }
         DebugLogger.shared.debug("Audio engine retired (\(reason))", source: "ASRService")
     }
@@ -2315,10 +2322,17 @@ final class ASRService: ObservableObject {
     private func registerEngineConfigurationChangeObserver() {
         guard self.engineConfigurationChangeObserver == nil else { return }
 
+        // queue: nil (synchronous delivery on the posting thread) is load-bearing:
+        // AVAudioEngine posts this notification from its internal serial queue, and
+        // NotificationCenter blocks a post until queued observers finish. With
+        // queue: .main that wait can never end when the main thread is itself
+        // blocked on the engine's queue (dealloc/stop during retirement) — a
+        // permanent deadlock (#542). The body only hops to the main actor, which
+        // is safe from any thread.
         self.engineConfigurationChangeObserver = NotificationCenter.default.addObserver(
             forName: .AVAudioEngineConfigurationChange,
             object: nil,
-            queue: .main
+            queue: nil
         ) { [weak self] notification in
             guard let changedEngine = notification.object as? AVAudioEngine else { return }
             Task { @MainActor [weak self, weak changedEngine] in
@@ -3549,6 +3563,25 @@ private extension ASRService {
     func stopStreamingTimer() {
         self.streamingTask?.cancel()
         self.streamingTask = nil
+    }
+}
+
+// MARK: - Audio engine retirement
+
+/// Carries the final strong reference to a retired audio engine so the release —
+/// and `-[AVAudioEngine dealloc]`, which blocks on the engine's internal serial
+/// queue — always happens on the drain queue, never on the main thread (#542).
+/// `@unchecked Sendable`: created on the main thread, then handed off and touched
+/// exactly once by the draining block; the dispatch provides the ordering.
+private final nonisolated class RetiredAudioEngineReference: @unchecked Sendable {
+    private var engine: AnyObject?
+
+    init(_ engine: AnyObject?) {
+        self.engine = engine
+    }
+
+    func drain() {
+        self.engine = nil
     }
 }
 
--- a/Sources/Fluid/Services/ASRService.swift
+++ b/Sources/Fluid/Services/ASRService.swift
@@ -613,7 +613,7 @@ final class ASRService: ObservableObject {
         if self.engineStorage != nil {
             let retired = RetiredAudioEngineReference(self.engineStorage)
             self.engineStorage = nil
-            DispatchQueue.global(qos: .utility).async { retired.drain() }
+            retired.scheduleRelease()
         }
         DebugLogger.shared.debug("Audio engine retired (\(reason))", source: "ASRService")
     }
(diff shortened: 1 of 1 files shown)


altic-dev:
@greptileai

PR #450, opened 2026-06-28
feat(menu-bar): add copy last transcript action

matthewrball:
## Summary

Adds a `Copy Last Transcript` item directly under the menu bar status row (`Ready to Record...`) so users can copy the most recent saved dictation without opening the main window or overlay actions menu.

## What changed

- Adds the menu item to `MenuBarManager` immediately below the status item.
- Reuses the existing transcription history and clipboard service.
- Centralizes the latest-copy text selection so both the overlay action and menu bar action prefer processed text, then fall back to raw text.
- Adds focused tests for the clipboard text selection behavior.

## Validation

- `git diff --check`
- `plutil -lint Fluid.xcodeproj/project.pbxproj Info.plist Fluid.entitlements`
- `xcodebuild test -project Fluid.xcodeproj -scheme Fluid -destination 'platform=macOS,arch=arm64' -only-testing:FluidDictationIntegrationTests/DictationE2ETests/testTranscriptionHistoryEntryClipboardTextPrefersProcessedText -only-testing:FluidDictationIntegrationTests/DictationE2ETests/testTranscriptionHistoryEntryClipboardTextFallsBackToRawText -only-testing:FluidDictationIntegrationTests/DictationE2ETests/testTranscriptionHistoryEntryClipboardTextSkipsEmptyText CODE_SIGN_IDENTITY='' CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO`
- `xcodebuild build -project Fluid.xcodeproj -scheme Fluid -configuration Debug -destination 'platform=macOS,arch=arm64' CODE_SIGN_IDENTITY='' CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO`

PR #568, opened 2026-07-10
feat(history): copy entries via double-click/⌘C/menu + keyboard navigation

### LINKED ISSUE #566: History: copy entry via double-click and ⌘C (keyboard) in Activity → History
### Feature request

In **Activity → History**, add two faster ways to copy a transcription entry to the clipboard, alongside the ones that already exist (the **Copy** button in the detail pane and the right-click **Copy Text** context menu):

1. **Double-click a row** in the history list → copies that entry's text to the clipboard.
2. **⌘C (Cmd+C)** while a row is selected → copies the selected entry's text.

### Why

The copy button and context menu both work, but they're an extra step/aim. Double-click and ⌘C are the muscle-memory gestures for "grab this text" everywhere else on macOS (Finder, Mail, Notes, most lists). For a dictation app where the whole point is getting text out quickly, making the history list copyable with the standard gestures removes friction. I'd use this constantly myself.

### Proposed behavior (details)

- **Double-click** selects the row (if not already) and copies its text — same text the existing **Copy** button uses (`processedText`, i.e. `TranscriptionHistoryEntry.clipboardText`).
- **⌘C** copies the currently-selected entry. No-op when nothing is selected.
- Both reuse the existing `copyToClipboard(_:)` helper — no change to *what* gets copied, only new triggers for it.
- A brief, subtle "Copied" confirmation would be a nice touch since double-click/⌘C give no other signal, but that's optional.

### Notes

- The history list is currently a custom `ScrollView` + `LazyVStack` of `Button` rows (not a `List(selection:)`), so ⌘C would be wired via a focus/key path rather than a list selection binding. Happy to implement it consistently with the existing code.
- I searched open/closed issues and PRs — this doesn't appear to be tracked yet. #450 added "copy last transcript" to the **menu bar**, which is adjacent but different (that's the last transcript, not an arbitrary selected/double-clicked row in History).

I'm happy to open a PR for this if it sounds good and fits the roadmap — just wanted to file the issue first per the contributing guidelines.

---

_Lineage: this builds on #3 (the original clipboard feature wish) and #450 (which added Copy Last Transcript to the menu bar). This issue extends the same idea into the History list._

### LINKED ISSUE #567: Activity → History: keyboard support for the entry list (↑/↓ navigation + ⌘C to copy)
### Feature request

Make the **Activity → History** entry list keyboard-navigable:

1. **↑ / ↓ move the selection within the History list** when the list is focused (click into it, then arrow up/down to move between entries — today the arrows move the *sidebar* selection instead: Stats ↔ History ↔ File Transcription).
2. **⌘C copies the selected entry** (respecting the search field, so it never shadows text copy there). Return/Enter could also copy or open.

### Why

Once you're browsing history, reaching for the mouse to change rows is friction. Arrow-key navigation + ⌘C is the standard macOS list interaction (Finder, Mail, Notes) and pairs naturally with the copy gestures.

### Root cause (why it doesn't work today)

The history entry list is a custom `ScrollView` + `LazyVStack` of `.buttonStyle(.plain)` `Button` rows with a manual `@State selectedEntryID`, not a focusable `List(selection:)` / `Table`. Because the list never becomes the keyboard first responder:

- arrow keys are handled by the sidebar's navigation instead of the list, and
- the standard **Copy command never reaches the list**, so a `.onCopyCommand` on it doesn't fire — which is why ⌘C currently does nothing for a selected history entry.

### Possible approaches

- Make the list container focusable and handle `.onMoveCommand` / `.onKeyPress` for ↑/↓ plus `.onCopyCommand` for ⌘C, keeping the current custom row visuals; or
- Migrate the custom list to a `List(selection:)`, which gives arrow-key navigation and the Copy command "for free" (at the cost of a larger visual/layout change to reconcile with the existing row styling).

### Relationship to #566

Split out from #566, which adds **double-click**, **right-click**, and **Copy-button** copying to the History list (with a small "Copied" toast near the cursor) — all of which work without keyboard focus. ⌘C was attempted in #566 via `.onCopyCommand` but couldn't fire because of the focus issue above, so it's tracked here to be implemented properly alongside arrow-key navigation. Happy to take this on as a follow-up PR.

Fe2-O3:
## Description

Adds faster copy affordances to **Activity → History** and makes the entry list keyboard-navigable:

- **Double-click** any entry → copies its transcription.
- **⌘C** → copies the currently-selected entry.
- **↑ / ↓** → move the selection *within the list* once it's focused (previously the arrows drove the sidebar selection).
- All four copy paths — double-click, ⌘C, the **Copy** button, and the right-click **Copy** menu — now share a single, consistent **"Copied"** confirmation: a small toast that appears next to the cursor and fades out, so the feedback shows up where you're already looking.

No change to *what* gets copied — every path reuses the existing `TranscriptionHistoryEntry.clipboardText` selection (processed-then-raw) centralized in #450, so gestures, keyboard, menu, and button all place identical text on the pasteboard.

### Why

The **Copy** button and right-click menu already work, but double-click and ⌘C are the muscle-memory gestures for "grab this text" everywhere else on macOS (Finder, Mail, Notes). For a dictation app whose whole point is getting text out quickly, wiring the standard gestures + arrow-key navigation into History removes friction.

### Implementation notes

- The history list was a custom `ScrollView` + `LazyVStack` of plain `Button` rows with a manual `@State` selection, so it never became the keyboard first responder — which is exactly why ⌘C and the arrow keys didn't reach it. The list container is now `.focusable()` and takes focus on row click, handling `.onMoveCommand` (↑/↓, with the selected row scrolled into view via `ScrollViewReader`) and `.onCopyCommand` (⌘C).
- Copy/selection logic is extracted into a small, UI-free `HistoryCopy` enum (`text(for:)`, `nextIndex(current:count:moveUp:)`) so it's unit-testable without a running app.
- The confirmation is a borderless, **non-activating** `NSPanel` at the cursor, so a subsequent ⌘V still targets the frontmost app.

## Type of Change
- [x] ✨ New feature

Closes #566
Closes #567

(Lineage: extends the clipboard idea from #3 and the menu-bar "Copy Last Transcript" from #450 into the History list.)

## Testing
- [ ] Tested on Intel Mac
- [x] Tested on macOS version: **27.0 (Apple Silicon, M1 Max)**
- [x] Ran linter locally: `swiftlint --strict --config .swiftlint.yml` on the changed files → **0 violations**
- [x] Ran formatter locally: `swiftformat --config .swiftformat` on the changed files → **0 require formatting**
- [x] Ran tests locally: 10 `HistoryCopy` unit tests pass — `** TEST SUCCEEDED **`

## Screenshots / Video

**"Copied" toast on a selected row (double-click / ⌘C):**

PR568-1-row-toast

**Same unified toast from the Copy button:**

PR568-2-copy-button

## Notes

- The new unit tests live inside the existing `DictationE2ETests.swift` rather than a standalone file: the test target uses explicit file membership (not folder-sync), so a new standalone test file wouldn't be compiled into the target.
- Happy to tweak the toast styling/timing, or to split the keyboard-navigation piece (#567) back into its own PR if you'd rather review them separately.

**Personal note:** I'm running a build with this now, and it's already made grabbing past transcriptions noticeably easier for me. I'm dealing with an injury that makes extra mouse-aiming and keystrokes costly, so double-click / ⌘C / arrow-key access to History is a real day-to-day accessibility win — which is a big part of why I wanted to contribute it back rather than just keep a local patch. Thank you for building FluidVoice and keeping it open source. 🙏


Fe2-O3:
Re: the copy-text consistency feedback from the automated review — I've routed all the "copy this transcription" paths (double-click, ⌘C, the detail-pane **Copy** button, and the context-menu **Copy Text**) through the existing `clipboardText` selection introduced in #450 — the same text the **Copy Last Transcript** menu action already copies. Previously the History Copy button copied raw `processedText`, which didn't trim and, for a transcription that wasn't AI-processed, copied an **empty string** (empty `processedText`, non-empty `rawText`) — so Copy on a raw transcription put nothing on the clipboard. Routing through `clipboardText` (trimmed, processed-then-raw) fixes that and makes every copy affordance in the app agree on what "copy this entry" means. **Copy Raw** / **Copy Both** are unchanged.

Also addressed the other review notes in this push: guarded ⌘C so it no longer copies the first row when nothing is explicitly selected, fixed the same nil-selection fallback in arrow-key navigation (the first ↑/↓ now lands on the first row instead of skipping it), hardened the toast's dismiss so a fresh copy firing mid-fade isn't hidden, and dropped the unused `UUID` parameter.

Fe2-O3:
Pushed `4074066` — the copy **buttons** now confirm inline instead of via the toast, per @grohith327's feedback in #566.

- The detail-pane buttons (**Copy** / **Copy AI**, **Raw**, **Both**) briefly become a checkmark + **Copied**, matching the shadcn `use-copy-to-clipboard` pattern you linked, so the feedback lands on the control you actually pressed.
- The cursor toast stays only on the paths that have no button to morph: **double-click**, **⌘C**, and the **right-click menu** (which dismisses the moment it's clicked). Without it, those actions would be silent.
- The copied state is scoped to its entry id, so the confirmation can't leak onto a different entry when the selection changes.

Also already in this PR from the earlier review round: all copy paths harmonized on the existing `clipboardText` selection (#450), ⌘C guarded when nothing is explicitly selected, the same nil-selection fallback fixed in arrow-key navigation, and the toast's dismiss hardened against a fresh copy firing mid-fade.

The substantive discussion of the remaining points (double-click, ⌘C) is over on #566 — happy to continue there or in a Discussion, whichever you prefer.

PR #590, opened 2026-07-11
Fix VoiceOver: hide recording overlays and pass through system hotkeys

Alshekhi:
## Description
Two accessibility fixes for VoiceOver users:

- **`GlobalHotkeyManager`** — a fast-path in the CGEvent tap that passes the macOS system shortcuts **Cmd+Tab** and **Cmd+Space** straight through, so the tap doesn't add latency to app/space switching while the dictation hotkey is active. This is especially noticeable with VoiceOver, whose own event handling is latency-sensitive. The fast-path is skipped when the user has bound one of Fluid's own shortcuts to the same chord, and it preserves the modifier-only bookkeeping so a system chord pressed mid-hold isn't mistaken for a clean tap.
- **`BottomOverlayView`** — `setAccessibilityElement(false)` + `setAccessibilityRole(.unknown)` on all four recording-overlay panels, and `.accessibilityHidden(true)` on the SwiftUI body, so VoiceOver doesn't announce the overlays or shift focus to them.

Both match behavior the repo's `CLAUDE.md` already documents as intended (overlay panels must hide from VoiceOver; the event tap must pass system shortcuts through immediately).

## Type of Change
- [x] 🐞 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 🧹 Chore
- [ ] 📝 Documentation update

Addresses #401 (*VoiceOver: dictation overlay steals focus and Cmd+Tab becomes sluggish*).

This PR fixes the two root causes in **this** repo:
1. `BottomOverlayView` panels being discoverable by VoiceOver (overlay focus stealing).
2. Cmd+Tab / Cmd+Space sluggishness from the CGEvent tap.

The third root cause noted in #401 — `DynamicNotchKit`'s `DynamicNotchPanel` overriding `canBecomeKey`/`canBecomeMain` — lives in the separate DynamicNotchKit dependency and is out of scope here; it only affects the non-default notch overlay and would need a fix in that repo. This PR covers the default **bottom** overlay path, so #401 is intentionally *addressed*, not auto-closed.

## Testing
- [ ] Tested on Intel Mac
- [x] Tested on macOS version: 26.5.2 (Apple Silicon)
- [ ] Ran tests locally:

Manually verified: pressing the dictation hotkey with VoiceOver active no longer delays Cmd+Tab / Cmd+Space; the recording overlays are no longer announced by or focusable in VoiceOver.

## Screenshots / Video

The overlay change only sets accessibility attributes (`setAccessibilityElement`, `setAccessibilityRole`, `.accessibilityHidden`) — it does not alter rendering in any way.

## Notes
Tested locally against the current `main`. The hotkey fast-path is limited to Cmd+Tab and Cmd+Space (both layout-stable hardware keycodes) to avoid the ANSI/ISO ambiguity of the backtick key.

### DIFF
--- a/Sources/Fluid/Services/GlobalHotkeyManager.swift
+++ b/Sources/Fluid/Services/GlobalHotkeyManager.swift
@@ -592,6 +592,19 @@ final class GlobalHotkeyManager: NSObject {
             return tapRecoveryResult
         }
 
+        // Fast path: pass through system shortcuts (Cmd+Tab, Cmd+Shift+Tab, Cmd+Space, etc.)
+        // to avoid adding latency to app switching, especially with VoiceOver.
+        if type == .keyDown || type == .keyUp {
+            let flags = event.flags
+            if flags.contains(.maskCommand) {
+                let kc = UInt16(event.getIntegerValueField(.keyboardEventKeycode))
+                // Tab(48), Space(49), Backtick(50 — Cmd+` for window cycling)
+                if kc == 48 || kc == 49 || kc == 50 {
+                    return Unmanaged.passUnretained(event)
+                }
+            }
+        }
+
         if self.isShortcutCaptureActiveProvider?() ?? false {
             self.resetModifierOnlyShortcutTracking()
             return Unmanaged.passUnretained(event)
--- a/Sources/Fluid/Services/GlobalHotkeyManager.swift
+++ b/Sources/Fluid/Services/GlobalHotkeyManager.swift
@@ -556,6 +556,11 @@ final class GlobalHotkeyManager: NSObject {
         if self.promptShortcutAssignments.contains(where: { $0.shortcut.matches(keyCode: keyCode, modifiers: modifiers) }) {
             return true
         }
+        if self.promptModeShortcutEnabled,
+           self.promptModeShortcut.matches(keyCode: keyCode, modifiers: modifiers)
+        {
+            return true
+        }
         if self.commandModeShortcutEnabled,
            let commandModeShortcut = self.commandModeShortcut,
            commandModeShortcut.matches(keyCode: keyCode, modifiers: modifiers)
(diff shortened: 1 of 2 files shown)


Alshekhi:
Thanks for the automated reviews — I've addressed every point:

- **Fast-path could swallow configured shortcuts** (Greptile P1 / Codex P2): `handleKeyEvent`'s fast-path now checks `eventMatchesAnyConfiguredShortcut(...)` before passing an event through, covering cancel, paste-last, prompt assignments, prompt-mode, command, rewrite, and primary shortcuts. A Fluid binding on Cmd+Tab/Space is no longer silently discarded.
- **Modifier-only bookkeeping skipped** (Codex P2): `markOtherInputDuringModifierOnly()` is now called before the pass-through, so a system chord pressed during a modifier-only shortcut isn't mistaken for a clean tap on release.
- **Keycode 50 is ANSI-only** (Greptile P2): dropped it. The fast-path now handles only Tab(48) and Space(49), which are layout-stable hardware keycodes; Cmd+` falls through to normal handling.
- Also extracted the `CGEventFlags` → `NSEvent.ModifierFlags` mapping into `modifierFlags(from:)` so the tap and the matcher share one source of truth.

**Scope:** this PR addresses root causes 1 and 3 of #401 (bottom-overlay accessibility + event-tap latency). Root cause 2 (DynamicNotchKit's `DynamicNotchPanel` overriding `canBecomeKey`/`canBecomeMain`) lives in the DynamicNotchKit dependency and only affects the non-default notch overlay, so it's out of scope here and tracked in #401.

Verified locally on Apple Silicon, macOS 26.5.2, with VoiceOver enabled: Cmd+Tab/Cmd+Space are no longer sluggish and the recording overlays are no longer announced by or focusable in VoiceOver.

PR #650, opened 2026-07-18
Fix: Escape (cancel) ignored during push-to-talk dictation

sw1pp3r:
## Problem

When the primary dictation shortcut is a **modifier key** (e.g. Right ⌘, the default is Right ⌥) and the hotkey mode is **Hold** or **Automatic**, the activation modifier is still physically held while recording (push-to-talk). Pressing the **cancel shortcut (Escape by default)** mid-hold does nothing — recording keeps running and is never discarded.

### Root cause

Every cancel check goes through `HotkeyShortcut.matches`, which requires an **exact** modifier match:

```swift
return keyCode == self.keyCode
    && modifiers.intersection(relevantModifierMask) == self.relevantModifierFlags
```

While push-to-talk holds the activation modifier, the Escape `keyDown` event reports that modifier on `event.flags`. So for a bare-Escape cancel shortcut (`relevantModifierFlags == []`):

```
matches(53, [.command])  →  53 == 53 ✓  &&  [.command] == [] ✗  →  false
```

The cancel key is silently ignored. If you instead **tap** the shortcut to toggle (modifier released), Escape works — which is why it's config-dependent. Affects the global event-tap path in `GlobalHotkeyManager` (the authoritative one that consumes the event first).

## Fix

Add `HotkeyShortcut.matchesAllowingHeldModifiers`, which still requires the shortcut's **own** modifiers but tolerates *extra* modifiers held on top. Use it in the cancel path **only while recording is active**, so cancel-key chords aren't over-consumed the rest of the time:

```swift
let cancelShortcutPressed = cancelShortcut.matches(keyCode: keyCode, modifiers: eventModifiers)
    || (self.asrService.isRunning
        && cancelShortcut.matchesAllowingHeldModifiers(keyCode: keyCode, modifiers: eventModifiers))
```

Behaviour preserved:
- Bare Escape still cancels as before.
- A cancel shortcut that itself has modifiers keeps requiring them (⌘+Escape does **not** fire on bare Escape).
- Outside of active recording, matching is unchanged (no extra event consumption).

## Tests

Added `testCancelShortcutMatchesWhilePushToTalkModifierIsHeld` covering the push-to-talk repro and the no-regression cases (bare Escape, wrong key, modifier-bearing cancel shortcut).

Also verified by compiling `HotkeyShortcut.swift` standalone against a probe harness — all assertions pass, including that the pre-fix `matches` reproduces the bug.

### DIFF
--- a/Sources/Fluid/Models/HotkeyShortcut.swift
+++ b/Sources/Fluid/Models/HotkeyShortcut.swift
@@ -282,6 +282,17 @@ extension HotkeyShortcut {
         return keyCode == self.keyCode && modifiers.intersection(Self.relevantModifierMask) == self.relevantModifierFlags
     }
 
+    /// Matches this shortcut's key while tolerating *extra* modifiers that are held down
+    /// but are not part of the shortcut. The shortcut's own modifiers (if any) must still
+    /// all be present. This lets a modifier-less cancel key (e.g. Escape) fire while a
+    /// push-to-talk activation modifier (e.g. Right ⌘) is still being held — the exact
+    /// `matches(_:_:)` check fails there because the held modifier is reported on the event.
+    func matchesAllowingHeldModifiers(keyCode: UInt16, modifiers: NSEvent.ModifierFlags) -> Bool {
+        guard !self.isMouseShortcut else { return false }
+        let activeModifiers = modifiers.intersection(Self.relevantModifierMask)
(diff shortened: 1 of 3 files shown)


grohith327:
Hey @sw1pp3r thanks, this looks like a useful change. Can you update the PR with the template and a recording ?

KnifeFed:
Would love to have this fixed. It has never worked.

Issue #592, opened 2026-07-12
[🐞 BUG] AirPods/Bluetooth dictation degraded: capture uses stale stream format and reports ready before the Bluetooth mic is live


rohithgoud30:
**Describe the bug**
When AirPods (or another Bluetooth headset) are the active input device, dictation is noticeably worse than with the built-in microphone: transcriptions come out garbled or time-stretched, the first words of a dictation are often missing, and overall accuracy drops. The built-in microphone works correctly on the same machine.

**Root causes (verified in the code)**
1. The direct Core Audio capture backend (enabled by default) reads the input stream's virtual format once in `fv_core_audio_capture_create` and never revalidates it. When the AirPods Bluetooth route switches from A2DP to HFP as input IO starts, the cached sample rate and byte layout go stale, and every packet is decoded and resampled with the wrong parameters, producing stretched/garbled audio that wrecks transcription.
2. There are no listeners for `kAudioStreamPropertyVirtualFormat` or `kAudioDevicePropertyNominalSampleRate`, so a mid-session format change is never detected and capture is never rebuilt.
3. `onCaptureStarted` fires immediately after `AudioDeviceStart` returns, but a Bluetooth mic can take 1 to 2 seconds to deliver its first packet, and audio timestamped before recording start is trimmed, so words spoken during HFP negotiation are silently lost. The UI tells the user to speak before the mic is actually live.
4. The 16 kHz conversion for ASR uses stateful linear interpolation with no anti-alias filter, folding content above 8 kHz into the speech band.

**To Reproduce**
1. Connect AirPods and let macOS make them the default input device.
2. Start dictation and begin speaking immediately.
3. Compare the transcript with the same sentence dictated via the built-in MacBook microphone.

**Expected behavior**
Dictation through AirPods should be reliable: no garbled/stretched audio, no missing first words, and accuracy close to the built-in microphone (within the physical limits of the Bluetooth HFP microphone route).

**Environment:**
 - macOS Version: 26.5.2 (25F84)
 - App Version: built from `main` @ 690b26c
 - Architecture: Apple Silicon (`arm64`)
 - Audio devices observed: AirPods (Bluetooth) input/output; MacBook Pro Microphone built-in

**Additional context**
- #452 documents the audible playback dip caused by the same A2DP to HFP route transition; this issue covers the capture-side fallout of that transition (stale format, lost first words, resampling quality).
[...]

Issue #477, opened 2026-06-29
[🐞 BUG] Start overlay/chime can indicate recording before audio capture is ready


freshyjmp:
**Describe the bug**
The recording overlay and start chime can indicate that dictation is ready before audio capture is actually stable. When I start speaking immediately after the hotkey/UI cue, the first words are sometimes missing from the transcript even though the rest of the dictation is accurate.

This is especially confusing because the black recording overlay appears immediately, which suggests the app is already listening. In practice, debug logs show `AVAudioEngine.start()` returning before a startup `AVAudioEngineConfigurationChange` route recovery finishes and before audio samples have accumulated.

**To Reproduce**
1. Enable dictation start sounds.
2. Trigger dictation with the global hotkey.
3. Start speaking as soon as the overlay/chime indicates recording.
4. Observe that the first words can be dropped, while later speech is transcribed correctly.

**Expected behavior**
The app should only play the start chime once capture is actually ready to receive audio, or the UI should show a distinct "starting" state until the capture graph is stable. The overlay/chime should not imply that speech is being captured while startup route recovery is still pending.

**Screenshots**
Not applicable.

**Environment:**
 - macOS Version: macOS 26.5.1 (25F80)
 - App Version: 1.6.1, installed via Homebrew cask
 - Architecture: Apple Silicon

**Additional context**
Observed debug timings from local testing:

- The recording overlay is shown immediately when dictation starts.
- `asr_start_return` has been observed around 460-662 ms after `asr_start_call`.
- A startup `AVAudioEngineConfigurationChange` can arrive after `asr_start_return`.
- Route recovery then completes roughly 140-170 ms later in tested sessions.
- Waiting for route recovery to be idle, a short stability delay, and at least a small captured sample buffer moved the start chime to the point where capture was actually ready.
- In local patched runs, the capture-ready cue fired roughly 513-932 ms after `asr_start_return`, with the start sound playing immediately afterward.

The main UX issue is that the current UI gives a false-ready signal. There may also be follow-up performance opportunities around startup graph construction, output node/device routing, or short-lived engine reuse, but a conservative first fix is to make the audible cue reflect capture readiness.

**Crash Logs**
No crash.

altic-dev:
We wrote a new pipeline for 1.6.2. And I think that will sort this problem out completely . Please give it a try and let me know what you think. Closing this issue for now, please open it up if it still exists. But most likely it's gone. Appreciate you bringing it up, a really important change :)

Issue #236, opened 2026-03-29
[🐞 BUG] Dictation activation delay causes first words to be dropped


reza-repo:
## Bug Description

When activating dictation via hotkey, there is a significant delay (~500ms–1s+) before the microphone actually starts capturing audio. Any words spoken during this startup window are permanently lost, as there is no mechanism to retroactively capture them.

This is especially noticeable when you start speaking immediately after pressing the hotkey, which is the natural user behavior.

## Steps to Reproduce

1. Launch FluidVoice with Parakeet (or any model)
2. Press the dictation hotkey
3. Immediately begin speaking a sentence (e.g., "Hello this is a test")
4. Observe the transcription output

**Expected:** Full sentence is captured — "Hello this is a test"
**Actual:** First 1 word is dropped — "This is a test"

**The following is done by Opus as why this is happening:**

## Root Cause Analysis

The activation flow in `ASRService.start()` runs multiple sequential blocking steps **before** installing the audio tap:

1. **Hotkey → `start()` dispatch** (~200ms) — `Task { @MainActor }` dispatch plus ContentView callback overhead
2. **Start sound plays synchronously** — before `ASRService.start()` is even called
3. **`configureSession()`** (~85–275ms) — forces `AVAudioEngine` input node instantiation
4. **`startEngine()`** (~150–320ms) — device binding + `prepare()` + `engine.start()`
5. **`setupEngineTap()`** (~1ms) — mic finally starts capturing here
6. **1-second minimum audio accumulation** — `processStreamingChunk()` requires 16,000 samples (1s at 16kHz) before attempting transcription

The core issue is that **audio capture does not begin until all setup is complete**, and `ThreadSafeAudioBuffer` is cleared on every `start()` call with no circular/ring buffer to preserve earlier audio.

Additionally, `stop()` deallocates the `AVAudioEngine` entirely, forcing a full rebuild on every subsequent `start()`.

## Related Issues

- #188 — ~5 second delay traced to `MediaRemoteAdapter` blocking main thread
- #98 — ~1 second delay acknowledged by developer, partially reduced by ~150ms

### 1. Always-on mic with a rolling circular buffer
Keep a ~2–3 second ring buffer of audio running in the background. On activation, prepend the buffered audio to the recording session. This way words spoken during startup are preserved.

### 2. Keep `AVAudioEngine` alive between sessions
[...]

PR #701, opened 2026-07-23
Delay dictation overlay until audio capture is ready

supfeer:
## Description

Starts audio capture before presenting the dictation UI so the beginning of a spoken phrase is retained.

This PR:

- marks capture ready only after the first PCM buffer is accepted
- waits 500 ms of active recording before showing the recording overlay
- plays the configured dictation start sound when the overlay appears
- cancels delayed presentation when capture stops or fails
- keeps app-context capture and AI prewarming on the first-audio boundary

## Type of Change
- [x] 🐞 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 🧹 Chore
- [ ] 📝 Documentation update

Related to #477, #236, and #592.

## Testing
- [ ] Tested on Intel Mac
- [x] Tested on macOS version: 26.5.2
- [ ] Ran tests locally:

Additional validation:

- Signed Debug build completed successfully.
- Installed and launched locally with `com.FluidApp.local`.

## Screenshots / Video

The existing overlay appearance and layout are unchanged. This PR only changes when the existing overlay and start sound are presented relative to active audio capture.

## Notes

- Audio acquisition begins before the 500 ms presentation delay.
- The delay is measured from the first accepted PCM buffer, not from the hotkey event.

### DIFF
--- a/Sources/Fluid/ContentView.swift
+++ b/Sources/Fluid/ContentView.swift
@@ -3045,9 +3045,6 @@ struct ContentView: View {
         self.advanceOverlayLifecycle()
         self.setActiveRecordingMode(.dictate)
         let shouldShowDictationOverlay = !self.isRecordingForCommand && !self.isRecordingForRewrite
-        let shouldPlayStartSound = !self.isRecordingForCommand
-            && !self.isRecordingForRewrite
-            && self.asr.micStatus == .authorized
 
         // Ensure normal dictation mode is set (command/rewrite modes set their own)
         if shouldShowDictationOverlay {
@@ -3055,15 +3052,9 @@ struct ContentView: View {
         }
 
         Task {
-            if shouldPlayStartSound, !self.asr.isRunning {
-                TranscriptionSoundPlayer.shared.playStartSound()
-            }
             await self.asr.start(onCaptureStarted: {
                 self.captureRecordingContext()
                 self.prewarmPrivateAIDictationIfNeeded(for: .primary)
-                if shouldShowDictationOverlay {
-                    self.menuBarManager.showRecordingOverlayImmediately()
-                }
             })
             if !self.asr.isRunning {
                 self.menuBarManager.hideRecordingOverlayImmediately(reason: "asr_start_failed")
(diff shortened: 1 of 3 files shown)


altic-dev:
This unfortunately goes against the UX I am planning toward - I am still A/B Testing to see which one makes sense and how to animate etc but adding 500ms will ruin the experience. Even the audio only takes ~150ms to start listening. 

Thanks for this but I won't be able to merge this as it would break the UX in a v bad way!

For next time, please discuss about this before raising a PR to make sure we can align beforehand.

aliashov-fnddev:
The original problem was that audio capture and the UI were not synchronized. This description may sound simplistic, but the fix corrects the entire startup flow rather than merely adding a static delay.
The issue is especially frustrating with Bluetooth devices: the overlay indicates that FluidVoice is listening even though the microphone is not yet ready to deliver audio. The UI is now synchronized with the first actual PCM buffer.
The additional 500 ms delay is only a UX refinement. It lets the audio engine run briefly before prompting the user to speak, reducing the chance that speech onset overlaps with recognition startup or audio-channel switching artifacts and clicks.

### Previous flow
Hotkey → start sound → audio backend starts → isRunning = true → overlay appears → first PCM buffer arrives
### Problems:

- AVAudioEngine.start() or AudioDeviceStart() could return when the device was formally started but no real audio data had arrived yet.
- isRunning was set immediately after the backend start method returned.
- MenuBarManager showed the overlay based on isRunning.
- ContentView could also independently request the overlay to appear immediately.
- The start sound played before audio capture was actually ready.
- With **Bluetooth devices**, profile switching and first-packet delivery introduced additional delay.

As a result, the overlay and start sound told the user it was safe to speak before AudioCapturePipeline had received a single PCM buffer. If the user started speaking immediately, the beginning of the phrase could be lost.
### Correct flow
Hotkey → backend starts → first PCM buffer arrives → 500 ms of active recording → start sound + overlay

altic-dev:
Thanks for explanation. I understand the PR and the purpose ofc. 

This is more like a patch instead of solving the root problem which is time to capture first audio.

PR #503, opened 2026-07-01
Fix deleted modifier-only primary shortcuts staying active( #498 )

Yaomeng1749:
## Summary

Fixes #498.

Clears stale Primary Dictation hotkey press state when the primary shortcut list changes. This prevents a deleted modifier-only primary shortcut from keeping a pending hold/tap path active until app restart.

## Changes

- Track removed primary shortcuts in `updatePrimaryShortcuts(_:)`.
- Clear/cancel matching primary dictation runtime state for removed primary shortcuts.
- Keep event tap teardown as a force-clear path.
- Add a regression test for deleting a modifier-only primary shortcut while its hold start is pending.

## Testing

Local:

```bash
swiftformat --config .swiftformat Sources/Fluid/Services/GlobalHotkeyManager.swift Tests/FluidDictationIntegrationTests/HotkeyShortcutTests.swift
swiftformat --lint --config .swiftformat Sources/Fluid/Services/GlobalHotkeyManager.swift Tests/FluidDictationIntegrationTests/HotkeyShortcutTests.swift
git diff --check
xcrun swiftc -parse Sources/Fluid/Services/GlobalHotkeyManager.swift Tests/FluidDictationIntegrationTests/HotkeyShortcutTests.swift
swiftlint --strict --config .swiftlint.yml
```

GitHub Actions, same head `c2caabd26d8c2a50daa6f061e78449aa233d8d5e`:

- Passing fork CI: https://github.com/Yaomeng1749/FluidVoice/actions/runs/28551228981
- Passed: SwiftLint, dependency resolution, build, tests.

Upstream Actions are awaiting maintainer approval for the public-fork workflow.

Issue #498, opened 2026-07-01
[BUG] Deleted modifier-only primary shortcut remains active until app restart


arthurlee116:
### Describe the bug

A deleted modifier-only Primary Dictation shortcut can continue activating FluidVoice until the app is restarted.

This looks like a remaining variant of #470, but with modifier-only shortcuts. The persisted settings no longer include the deleted shortcut, yet the old physical modifier key can still trigger dictation in the running app.

### What happened

1. Primary Dictation Shortcut was set to Left Option.
2. Left Command was accidentally recorded as an additional Primary Dictation Shortcut.
3. Left Command was then deleted from FluidVoice settings.
4. After deletion, pressing Left Command still activated FluidVoice.
5. Restarting/quitting FluidVoice cleared the stale behavior.

### Local evidence

After deleting Left Command, FluidVoice preferences only showed Left Option (`keyCode: 58`):

```json
HotkeyShortcutKey: {"kind":"keyboard","keyCode":58,"modifierKeyCodes":[58],"modifierFlagsRawValue":0}
PrimaryDictationShortcuts: [{"kind":"keyboard","keyCode":58,"modifierKeyCodes":[58],"modifierFlagsRawValue":0}]
```

A separate CGEvent listener confirmed the physical Left Command key reports `keyCode=55`, not `58`. So the deleted Left Command shortcut was no longer present in persisted preferences, but it was still active in the running hotkey manager.

### Expected behavior

After deleting a Primary Dictation shortcut, the deleted modifier-only shortcut should stop triggering immediately, without requiring an app restart.

### Actual behavior

The deleted modifier-only shortcut can remain active until FluidVoice is restarted.

### Environment

- FluidVoice: 1.6.1
- macOS: Tahoe 26.x
- Keyboard event codes involved:
  - Left Command: `55`
  - Left Option: `58`
  - Right Option: `61`

### Possible source area

`GlobalHotkeyManager.updatePrimaryShortcuts(_:)` appears to only assign the new shortcut array:

```swift
func updatePrimaryShortcuts(_ newShortcuts: [HotkeyShortcut]) {
    self.primaryShortcuts = newShortcuts
    DebugLogger.shared.info("Updated transcription hotkeys", source: "GlobalHotkeyManager")
}
```

For modifier-only shortcuts, it may also need to clear modifier-only/active press state or reinitialize the tap when the shortcut set changes.

### Related

[...]

Issue #470, opened 2026-06-29
[🐞 BUG] Changing Primary Dictation shortcut doesn't remove/overwrite previous shortcut


knavet:
**Describe the bug**
With my primary dictation shortcut set as a SHIFT + ` key, i tried to change to my HyperKey + ` but after setting, both old and new shortcut still triggers the dictation until app restart.

**To Reproduce**
1. Go to 'Settings -> Global Shortcuts'
2. Click on 'change' under Primary Dictation Shortcuts section. Input the new shortcut pattern.
3. Try to activate both new/old shortcut, both triggers the dictation.

**Expected behavior**
New shortcut should have overwritten the old shortcut and only the new should work without an app restart.

**Screenshots**
If applicable, add screenshots.

**Environment:**
 - macOS Version: Tahoe 26.5.1
 - App Version: 1.6.1
 - Architecture: [Apple Silicon]

**Additional context**
-

**Crash Logs**
-

altic-dev:
This was fixed and seconary shortcut was replaced by multiple primary! 

Forgot to close this automatically.. thanks

PR #707, opened 2026-07-24
fix(hotkey): make Change replace and Remove unbind

dschnurbusch:
## Description

Changing or removing a Primary Dictation shortcut currently mutates SwiftUI state and relies on an `onChange` side effect to persist the list and refresh `GlobalHotkeyManager`. The shortcut-capture callback can bypass that refresh, leaving the manager's in-memory list stale: **Change** behaves like **Add**, and removed shortcuts continue firing until restart.

This PR routes **Change** and **Remove** through one persist-and-live-update method. It also permits removing the final primary shortcut, preserving an explicit unbound state instead of silently restoring the legacy shortcut. Fresh installs still receive the existing default when no primary-shortcut preference has ever been stored.

## Type of Change
- [x] 🐞 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 🧹 Chore
- [ ] 📝 Documentation update

Related to #470 and #498.

#474 contains the direct live-manager refresh for **Change**, but does not address **Remove** or an explicitly unbound primary shortcut. #503 covers a narrower active modifier-only press-state edge case; this PR addresses settings-to-live-manager synchronization and removal semantics observed on 1.6.5.

## Testing
- [ ] Tested on Intel Mac
- [ ] Ran tests locally:
- [x] Parsed all changed Swift files with `xcrun swiftc -parse`
- [x] Ran `git diff --check`
- [x] Fork CI passed SwiftLint, Xcode build, and tests: https://github.com/dschnurbusch/FluidVoice/actions/runs/30101354571

The local machine has Command Line Tools rather than full Xcode, so the full Xcode build/test suite was run on the same commit (`83568b8`) through the fork's macOS workflow.

## Screenshots / Video

Before: FluidVoice 1.6.5 shows **Remove** disabled when only one Primary Dictation shortcut remains. The submitted change enables removal of that final shortcut; once removed, the row disappears and **Add shortcut** remains available.

FluidVoice 1.6.5 Global Hotkey settings with disabled Remove button

- [ ] No UI/visual changes; screenshots/video are not applicable.

## Notes

The settings mutation now updates persisted state, normalized SwiftUI state, and the live hotkey manager in one call. The existing `onChange` remains as an idempotent fallback for other state writes.

The explicit empty list is distinguishable from a fresh install: missing preference data still falls back to the legacy/default shortcut, while stored `[]` remains unbound. The legacy single-shortcut value is left intact for backward-compatible backup/export consumers but is no longer allowed to resurrect an explicitly cleared primary list.

### DIFF
--- a/Sources/Fluid/ContentView.swift
+++ b/Sources/Fluid/ContentView.swift
@@ -1119,7 +1119,14 @@ struct ContentView: View {
                 shortcuts.append(shortcut)
             }
         }
-        self.primaryDictationShortcuts = shortcuts
+        self.updatePrimaryDictationShortcuts(shortcuts)
+    }
+
+    private func updatePrimaryDictationShortcuts(_ shortcuts: [HotkeyShortcut]) {
+        SettingsStore.shared.primaryDictationShortcuts = shortcuts
+        let storedShortcuts = SettingsStore.shared.primaryDictationShortcuts
+        self.primaryDictationShortcuts = storedShortcuts
+        self.hotkeyManager?.updatePrimaryShortcuts(storedShortcuts)
     }
 
     private func setShortcutTargetEnabled(_ enabled: Bool, for target: ShortcutRecordingTarget) {
@@ -1464,7 +1471,10 @@ struct ContentView: View {
             inputDevices: self.$inputDevices,
             outputDevices: self.$outputDevices,
             accessibilityEnabled: self.$accessibilityEnabled,
-            primaryDictationShortcuts: self.$primaryDictationShortcuts,
+            primaryDictationShortcuts: Binding(
+                get: { self.primaryDictationShortcuts },
+                set: { self.updatePrimaryDictationShortcuts($0) }
+            ),
             activeShortcutRecordingTarget: self.$activeShortcutRecordingTarget,
             shortcutRecordingMessage: self.$shortcutRecordingMessage,
             commandModeShortcut: self.$commandModeHotkeyShortcut,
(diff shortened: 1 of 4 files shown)


dschnurbusch:
Full fork CI passed on the same head commit (83568b8): SwiftLint, Xcode build, and tests all green. Run: https://github.com/dschnurbusch/FluidVoice/actions/runs/30101354571

altic-dev:
Thanks for the PR! 

Does this affect only the primary key or can we translate this to other keys and hotkey related areas?

dschnurbusch:
The concrete bug and code path in this PR are specific to **Primary Dictation**. Primary is the outlier because it is a list cached by `GlobalHotkeyManager`, and Settings UI mutations were relying on SwiftUI `onChange` to persist the list and refresh that cache. This PR gives Add/Change/Remove one explicit path that updates the store, normalized view state, and live manager together; the persistence change also lets an empty primary list mean intentionally unbound.

The same persist-and-live-update principle applies to the other hotkeys, but most already do the equivalent:

- Secondary Dictation, Command Mode, and Edit Mode explicitly persist replacements and call their corresponding manager update methods. Their enabled toggles also update the manager.
- Command Mode already supports Remove by setting the shortcut to `nil` and disabling it. Paste Last also supports Remove, and both Paste Last and Cancel are read directly from `SettingsStore` while handling events rather than from a stale manager cache.
- Prompt-specific shortcuts rebuild the manager assignments explicitly after a change.

So the pattern is reusable, but I would not directly broaden this PR: Remove/unbind for Secondary, Edit, or Cancel would be a separate UX/defaults decision, not the same stale-cache bug. A later cleanup could centralize each shortcut mutation behind a persist + runtime-sync API so the UI cannot accidentally update only one layer.

PR #711, opened 2026-07-25
fix(settings): clear orphaned per-prompt shortcuts on legacy retirement

### LINKED ISSUE #675: [BUG] Left Command key triggers voice detection despite Option Right being configured
### Describe the bug

The left Command key (Cmd) systematically activates voice detection/dictation, even though it is not configured in the settings.

My default shortcut for speech detection is set to Option Right. However, pressing the Left Command key triggers the speech recognition as well.

This completely breaks normal workflow on macOS. Since the Command key is essential for basic OS and application shortcuts (Cmd + C, Cmd + V, Cmd + Z, Cmd + Tab, Cmd + W, etc.), every single standard system shortcut inadvertently launches the dictation feature.

Because of this, the app is currently unusable for daily work.

### Reproduction steps

1. Open Fluid Voice settings and ensure the shortcut trigger is set to Option Right (or any key other than Left Command).
2. Press the Left Command key on the keyboard (or use any shortcut involving it, like Cmd + C).
3. Observe that voice detection/speech recognition is triggered.

### Expected behavior

Only the shortcut configured in the settings (Option Right) should trigger voice detection. The Left Command key should be completely ignored by Fluid Voice unless explicitly mapped.

### Actual behavior

The left Command key (Cmd) systematically activates voice detection/dictation, even though it is not configured in the settings.

### App Version

1.6.5

### macOS Version

maxOS 26.5.2

### Architecture

Apple Silicon

### Logs or crash report

```text

```

_No response_

shivanshsen7:
## Description

Retiring the legacy secondary-prompt-mode hotkey (`retireLegacySecondaryPromptShortcutIfNeeded()`) clears `PromptModeHotkeyShortcut` / `PromptModeSelectedPromptID`, but never touches `DictationPromptConfigurations` — a separate, still-active data store used by the current "Custom shortcut per AI Prompt profile" feature.

`dictationPromptShortcutAssignments()` turns every stored `.shortcut` in `DictationPromptConfigurations` into a live global hotkey unconditionally, with no dependency on `PromptModeShortcutEnabled` or either retirement flag. Any residual `.shortcut` left over from before the legacy hotkey was retired — most commonly a bare Command keypress under the `__default__` entry — keeps firing forever, even though nothing in Settings shows or lets you clear it.

This matches the two independent reports on #675: both show the exact same orphaned `keyCode: 55` (Command) under `__default__`, despite having different Primary Dictation shortcuts configured (Right Option / Right Control). That consistency across unrelated users is why I'm confident this is leftover legacy state rather than an intentional per-profile shortcut.

This PR clears any stale `.shortcut` bindings once, at the same retirement boundary that already governs `PromptModeHotkeyShortcut`, while preserving any real `providerID`/`modelName` overrides a user may have set on a prompt configuration.

## Type of Change
- [x] 🐞 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 🧹 Chore
- [ ] 📝 Documentation update

Fixes #675

## Testing
- [ ] Tested on Intel Mac
- [x] Tested on macOS version: 26.5.2
- [ ] Ran linter locally: `swiftlint --strict --config .swiftlint.yml Sources` (SourceKit crashes on this machine — only Command Line Tools installed, no full Xcode; see note below)
- [x] Ran formatter locally: `swiftformat --config .swiftformat Sources` — diffed `--lint` output against unmodified `main` for both touched files; this change introduces zero new formatting findings (45 pre-existing findings unrelated to this change, unchanged before/after)
- [ ] Ran tests locally (blocked — see note below)

**Note on local verification:** this machine only has Command Line Tools, not full Xcode, so `xcodebuild test` isn't available and `swiftlint` crashes on a missing `sourcekitdInProc` framework. I verified both changed files parse cleanly with `xcrun swiftc -parse`, and confirmed with `swiftformat --lint` that this diff adds no new style issues. I also manually reproduced the exact preference-file shape from #675 on my own affected install (`DictationPromptConfigurations.__default__.shortcut = {keyCode: 55, modifierKeyCodes: [55]}`, `PromptModeShortcutEnabled = false`) and applied the equivalent of this fix via `defaults delete` as a live workaround, which stopped Command from triggering dictation without disturbing the configured Primary Dictation shortcut. The new regression test reproduces that same data shape and asserts `dictationPromptShortcutAssignments()` — the source `GlobalHotkeyManager` reads for these hotkeys — ends up empty after cleanup. CI (`build.yml`, `macos-latest` with full Xcode) should be able to run the full suite including this test.

## Screenshots / Video

## Notes

`clearOrphanedDictationPromptShortcuts()` clears `.shortcut` on *every* entry in `DictationPromptConfigurations`, not just `__default__`, since the legacy single-shortcut system could plausibly have left residue under `__privateAI__` or a `profile:*` key too. The tradeoff: if anyone genuinely used the current "Custom shortcut" editor (Advanced AI Settings → prompt editor → "Custom shortcut") to bind a shortcut before updating past this migration, that binding would also be cleared. Given both #675 reports show the identical `keyCode: 55` artifact regardless of their actual configured Primary Dictation shortcut, I believe this is safe, but happy to narrow the scope (e.g. only `__default__`, or only shortcuts matching the known legacy default) if you'd prefer — you have visibility I don't into how many users use per-profile custom shortcuts today.

### DIFF
--- a/Sources/Fluid/Persistence/SettingsStore.swift
+++ b/Sources/Fluid/Persistence/SettingsStore.swift
@@ -3343,7 +3343,31 @@ final class SettingsStore: ObservableObject {
         self.defaults.set(true, forKey: Keys.secondaryDictationPromptOff)
         self.defaults.removeObject(forKey: Keys.promptModeHotkeyShortcut)
         self.defaults.removeObject(forKey: Keys.promptModeSelectedPromptID)
+        self.clearOrphanedDictationPromptShortcuts()
         self.defaults.set(true, forKey: Keys.legacySecondaryPromptShortcutRetired)
+    }
+
+    /// Before per-prompt "Custom shortcut" assignments existed, the single legacy secondary-prompt
+    /// hotkey (`PromptModeHotkeyShortcut` / `PromptModeSelectedPromptID`, cleared above) left a
+    /// residual `.shortcut` behind in `DictationPromptConfigurations` — most commonly a bare Command
+    /// keypress under the `__default__` entry. `dictationPromptShortcutAssignments()` treats any
+    /// stored `.shortcut` as an always-on global hotkey with no dependency on
+    /// `PromptModeShortcutEnabled`, so that residue kept firing dictation on every Cmd+key chord even
+    /// though the feature that created it was retired (altic-dev/FluidVoice#675). Provider/model
+    /// overrides are unaffected — only the shortcut binding is cleared, and only once, here, at the
+    /// same retirement boundary that already governs the legacy key.
+    func clearOrphanedDictationPromptShortcuts() {
+        var configurations = self.dictationPromptConfigurations
+        guard configurations.values.contains(where: { $0.shortcut != nil }) else { return }
+
+        for key in configurations.keys {
+            configurations[key]?.shortcut = nil
+        }
+        configurations = configurations.filter { _, configuration in
+            !configuration.providerID.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty ||
+                !configuration.modelName.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
+        }
+        self.dictationPromptConfigurations = configurations
     }
 
     private func normalizePromptSelectionsIfNeeded() {
--- a/Tests/FluidDictationIntegrationTests/HotkeyShortcutTests.swift
+++ b/Tests/FluidDictationIntegrationTests/HotkeyShortcutTests.swift
@@ -10,6 +10,7 @@ final class HotkeyShortcutTests: XCTestCase {
     private let pasteLastTranscriptionShortcutKey = "PasteLastTranscriptionHotkeyShortcut"
     private let pasteLastTranscriptionEnabledKey = "PasteLastTranscriptionShortcutEnabled"
     private let microphoneSelectionModeKey = "MicrophoneSelectionMode"
+    private let dictationPromptConfigurationsKey = "DictationPromptConfigurations"
     private let preferredInputDeviceUIDKey = "PreferredInputDeviceUID"
     private let systemInputDeviceUIDBeforeManualKey = "SystemInputDeviceUIDBeforeManual"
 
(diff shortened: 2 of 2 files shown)


postoso:
Had a close look at this against current main after #815 landed. The diagnosis and the plist evidence in #675 are useful. I found two coupled blockers in the implementation:

**The cleanup cannot run on the installs it targets.** `retireLegacySecondaryPromptShortcutIfNeeded()` returns when `LegacySecondaryPromptShortcutRetired` is already true. That flag has been set once on first launch since 1.6.3, and it is already true in the affected plist shared in #675. The new cleanup needs its own one-shot version key rather than reusing the retirement flag.

**Giving it a fresh key makes the current cleanup too broad.** `DictationPromptConfigurations` is the live store behind Advanced AI Settings > prompt profiles > Custom shortcut. Clearing `.shortcut` from every `__default__`, `__privateAI__`, and `profile:*` entry would erase legitimate user-created bindings. The migration should be restricted to an independently confirmed orphan signature and preserve every other prompt shortcut.

I would also pin down the writer before choosing that signature. The retirement commit does not write to `DictationPromptConfigurations`, that store arrived with the current per-prompt editor, and the old `PromptModeHotkeyShortcut` default was Right Shift (key code 60), not Command (55). If a live recorder path can still commit bare Command, a one-shot cleanup will not hold.

One smaller workflow note: because this is still a draft, the main build/lint/test workflow has not run yet. The branch conflict is limited to `HotkeyShortcutTests.swift` next to #815's new tests; `SettingsStore.swift` merges cleanly.

Happy to help with the rebase and run the full lint/build/test gate once the migration signature is narrowed, if that would be useful.

shivanshsen7:
I just added as draft coz i felt that it will be burden as its full ai generated and i thought it could help as i loved the app.
If its for me to work on it, i'd love to contribute back, or if its better that you take ownership feel free to.
Regardless i will rebase and work on ur feedback so changes are visible to you.

postoso:
whatever you want. I also use this app all the time and so I've started contributing too.

altic-dev:
should we check this out properly @postoso @shivanshsen7 ? I am looking to clear hotkey issues and this popped up adn was curious on what's the final decision on this.

postoso:
The repro and plist evidence were useful, but my final call is to close #711 as written and keep #675 open for a replacement fix.

The bug is still live on main. Stored prompt shortcuts are still registered without an enable check, and #815 only narrows the symptom: a bare Command tap still triggers, while hold and automatic modes start on the modifier press. #711 cannot remediate the reported affected state because its cleanup is inside `retireLegacySecondaryPromptShortcutIfNeeded()`, which returns once `LegacySecondaryPromptShortcutRetired` is already true. It also clears every prompt shortcut, including supported user-created bindings, and its test treats that loss as expected for `__privateAI__`.

On current main, I found no migration or automatic copy path from the legacy or Primary shortcut into `DictationPromptConfigurations`; the Custom shortcut editor can persist the exact reported `{keyCode: 55, modifierKeyCodes: [55]}` value, but the source of the affected users' records remains unresolved. This needs a replacement patch, not a rebase: any automatic remediation needs a fresh migration key and a maintainer-approved orphan signature that preserves every other prompt shortcut. The one product decision left is whether to treat the exact empty `__default__` Left Command record reported in #675 as orphaned despite the chance that it was intentionally recorded.

Issue #18, opened 2025-11-30
Add speaker diarization (speaker attribution) for multi-speaker conversations


akirill0v:
Hi, and thanks a lot for FluidVoice – it’s an awesome project!

Hi! I'd like to request a speaker diarization feature for FluidVoice. Currently, all transcribed text is attributed to a single speaker, which makes it difficult to read or use transcripts from meetings, interviews, and podcasts. It would be great if the app could automatically detect different voices and label them (e.g., "Speaker 1", "Speaker 2") in the live preview and history. This would significantly improve the utility of FluidVoice for recording multi-person conversations without requiring manual attribution. 

Thanks for the great tool!

altic-dev:
I will add it to the ask list and work on it soon for you and others!! Thanks for trying FV :)

benja0x40:
Hey, would be please to help and keen to start learning Swift to do so.

I did some successful tests on Mac with [pyannote/speaker-diarization-3.1](https://huggingface.co/pyannote/speaker-diarization-3.1) and [nvidia/parakeet-tdt-0.6b-v2](https://huggingface.co/nvidia/parakeet-tdt-0.6b-v2), though exclusively in Python. Also found an alternative model for native Mac diarization ([FluidInference/speaker-diarization-coreml](https://huggingface.co/FluidInference/speaker-diarization-coreml)) but didn't make any tests with it.

Below are other improvements I was considering to include in my Python pipeline.

**Pre Diarization-Transcription**
- Voice Activity Detection, based on [pipecat-ai/smart-turn-v2](https://huggingface.co/pipecat-ai/smart-turn-v2) model
- Sound Quality Improvement (e.g. noise reduction), postponed

**Post Diarization-Transcription**
- Multilingual Sentence Detection, based on [segment-any-text/sat-3l-sm](https://huggingface.co/segment-any-text/sat-3l-sm) or better
- Multilingual Punctuation, based on [oliverguhr/fullstop-punctuation-multilang-large](https://huggingface.co/oliverguhr/fullstop-punctuation-multilang-large)
- Multilingual Spelling Correction, based on [oliverguhr/spelling-correction-multilingual-base](https://huggingface.co/oliverguhr/spelling-correction-multilingual-base)

[...]

PR #766, opened 2026-08-03
Speaker labeling for file transcription, rebased onto 1.6.6, with a stereo channel fix

Erik5000:
## Description

This carries @postoso's speaker-labeling work from #378 forward: their three commits rebased onto current `main` (1.6.6) with authorship preserved, plus one bug fix and tests.

#378 has been open since June and now conflicts with `main`. The feature itself is sound — this is the same work, made mergeable again, with one defect fixed.

**Credit and precedence:** the feature is @postoso's. Commits 1–3 are theirs, unchanged. I offered them first refusal on [#378](https://github.com/altic-dev/FluidVoice/pull/378#issuecomment-5162946836) — to push this branch to their own fork, or have me open it here. That offer still stands: **if @postoso would rather carry it themselves, close this and I'll hand over the branch.** I opened it so the work isn't blocked on a reply.

### The bug fix

`MeetingTranscriptionService` had gained its own `resampleBuffer` helper. It built an `AVAudioConverter` but never set `converter.downmix`, so a multi-channel source was channel-mapped rather than mixed: **only the left channel was transcribed and the right one silently discarded.**

The standard file-transcription path *in the same file* already calls `AudioBufferConverter.monoSamples`, which sets `downmix` and is covered by `testStereoDownmixIncludesRightChannel` and `testStereoDownmixAndResampleIncludesRightChannel`. Only the speaker-labeled path used the second, subtly different copy — which is why it never surfaced in normal use.

Deleting the copy and delegating removes 75 lines and adds no new behaviour.

Found by feeding a stereo file with different speech on each channel: the transcript held only the left channel, and swapping the channels swapped the output. Stereo is common — in one real-world archive of 164 recordings, 72 were stereo.

**Known tradeoff:** mixing is the right default over dropping half the content, but it destroys an exact separation when the channels genuinely hold different speakers. Surfacing that to the user is worth doing separately; silently transcribing half the file is not.

### Tests

`SpeakerTurnMergingTests` pins down `mergeAdjacentTurns`: same-speaker merging across short gaps, the gap boundary, the 20-minute cap, and the consequence that matters — when clustering collapses to a single label, the merge welds question and answer into one segment.

## Type of Change
- [x] 🐞 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 🧹 Chore
- [ ] 📝 Documentation update

Addresses #18 (speaker diarization). Supersedes #378 — same work, rebased; @postoso remains the author of the feature commits.

## Testing

- [x] Tested on macOS version: 26.x (Darwin 25.5)
- [x] Ran linter locally: `swiftlint --strict --config .swiftlint.yml Sources` — 0 violations
- [x] Ran formatter locally: `swiftformat --config .swiftformat Sources` — this branch introduces no new findings; the ones that remain in these files are pre-existing on `main`
- [x] Ran tests locally: `xcodebuild test -scheme Fluid -destination 'platform=macOS,arch=arm64'` — **174 tests, 0 failures**
- [ ] Tested on Intel Mac — **not applicable**: the feature is `arch(arm64)`-gated and the toggle is hidden on Intel, unchanged from #378.

## Screenshots / Video

Captured against this branch. The audio is an 80-second public-domain excerpt from a LibriVox dramatic reading, deliberately not a real recording — a screenshot publishes whatever words are on screen.

**"Label speakers" toggle and speaker-count picker** (appear once a file is selected):

Label speakers toggle

**Speaker-labeled result with timestamps:**

Speaker labelled result

**Diarize-first pipeline in progress** — one ASR call per speaker turn:

Transcribing speaker segments

Two ASR artefacts are visible in the result and worth naming rather than hiding: the first segment ends with a hallucinated "Thank you." on trailing silence, and punctuation degrades toward the end of that long turn. Neither is introduced by this PR — both are existing recogniser behaviour on a long turn, unchanged by the speaker-labeling work. The speaker attribution itself is correct: the switch lands at 0:40, where the recording actually changes voice.

## Notes

The rebase had one conflict, in `MeetingTranscriptionService.swift`, resolved toward main's current structure — `main` has reworked the surrounding file since June. No behavioural change was made while resolving it.

A follow-up is ready that reports diarization quality rather than falling back silently, so a caller can distinguish "diarization found one speaker" from "diarization failed". Keeping it out of this PR to hold the diff to the feature plus the fix.


altic-dev:
@postoso  - was this your PR - the pilot one for diarization? trying to understand as there's too much here lmao. 

we can try to squeeze this into 1.6.7 or 1.6.8 based on how it goes and if we can get it stablized overall. Sorry it's been out for a while and appreciate the support in rebasing it etc. 

Happy to prioritize this! A lot of the users are asking for meeting transcription, and this is going to be a good pull.  @Erik5000

altic-dev:
@grohith327 Viz

postoso:
Yep, #378 is my original pilot PR. Erik opened this branch to rebase it and fix the stereo-audio path; I’ve since integrated those commits back into #378 with his authorship preserved, plus the follow-up fixes/tests, and all of its checks are green.

To keep review in one place, I’d treat #378 as the canonical PR and close #766 as the handoff/duplicate. And agreed on scope: this PR should stay focused on speaker-labeled file transcription. Broader meeting transcription can build on it separately for 1.6.7/1.6.8.

PR #795, opened 2026-08-05
Fix duplicate instances remaining after update

### LINKED ISSUE #794: Race condition in AppDelegate causes multiple app instances during LaunchServices reopen
When starting FluidVoice, if the SwiftUI main window takes too long to render (e.g., when the system is under heavy load or audio subsystems take a moment to initialize), the app's `AppDelegate` attempts to force the window open by requesting a LaunchServices reopen.

However, this reopen command spawns a completely new process instance in the background because it appears to be using `createsNewApplicationInstance = true` (or equivalent `open -n` logic) instead of just bringing the existing instance to the front. Because the original instance never actually terminates, the user ends up with duplicate `FluidVoice` processes running at the same time, causing conflicts with global hotkeys and microphone capture.

Logs showing the behavior:
```
[RUN] PID=44596 AppVersion=1.6.7 Build=18 OS=Version 15.7.7
[13:11:17.107] [DEBUG] [AppDelegate] Main window not ready during launch reveal retry
...
[13:11:17.632] [DEBUG] [AppDelegate] Main window not ready during launch reveal retry
[13:11:17.632] [INFO] [AppDelegate] Requesting LaunchServices reopen to create SwiftUI main window
[13:11:17.651] [INFO] [SimpleUpdater] SimpleUpdater: Successfully relaunched app, terminating old instance
[RUN] PID=44611 AppVersion=1.6.7 Build=18 OS=Version 15.7.7 (Build 24G720)
```
At this point, both `PID 44596` (the original) and `PID 44611` (the duplicate) remain running indefinitely.

**Suggested Fix:**
The LaunchServices configuration in `AppDelegate` should not use `setCreatesNewApplicationInstance: YES` when attempting to restore the SwiftUI window scene. Instead, it should just request the existing application to activate, or the retry logic for the SwiftUI lifecycle should be re-evaluated.

sfasano:
## Description
Replaced `NSApp.terminate(nil)` with `exit(0)` in `SimpleUpdater.swift`. This ensures the old instance forcefully quits after launching the new application instance, preventing a race condition where duplicate processes run simultaneously and fight over the global hotkey and microphone.

## Type of Change
- [x] 🐞 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 🧹 Chore
- [ ] 📝 Documentation update

Closes #794

## Testing
- [x] Tested on Intel Mac
- [x] Tested on macOS version: 15.7
- [ ] Ran tests locally:

## Screenshots / Video

## Notes
The `exit(0)` call prevents the application from lingering in the background if the runloop is blocked during an update.

### DIFF
--- a/Sources/Fluid/Services/SimpleUpdater.swift
+++ b/Sources/Fluid/Services/SimpleUpdater.swift
@@ -918,14 +918,14 @@ final class SimpleUpdater {
                 DebugLogger.shared.info("SimpleUpdater: Successfully relaunched app, terminating old instance", source: "SimpleUpdater")
                 // Give the new instance time to fully start before terminating
                 DispatchQueue.main.asyncAfter(deadline: .now() + 2.0) {
-                    NSApp.terminate(nil)
-                    
-                    // Fallback: forcefully terminate if AppKit gets stuck during or after termination.
+                    // Arm the fallback watchdog BEFORE calling terminate, because terminate blocks.
                     // The AppDelegate cleanup can take up to 16 seconds (8s for Private AI + 8s for ASR).
                     // We wait 20 seconds on a background queue to ensure it only fires if truly hung.
                     DispatchQueue.global().asyncAfter(deadline: .now() + 20.0) {
                         exit(0)
                     }
+                    
+                    NSApp.terminate(nil)
                 }
             }
         }
--- a/Sources/Fluid/Services/SimpleUpdater.swift
+++ b/Sources/Fluid/Services/SimpleUpdater.swift
@@ -920,7 +920,10 @@ final class SimpleUpdater {
                 DispatchQueue.main.asyncAfter(deadline: .now() + 2.0) {
                     NSApp.terminate(nil)
                     
-                    DispatchQueue.main.asyncAfter(deadline: .now() + 2.0) {
+                    // Fallback: forcefully terminate if AppKit gets stuck during or after termination.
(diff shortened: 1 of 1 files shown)


altic-dev:
Thanks for the PR. I'm not sure if killing it forcefully will remove the memory occupied by models that FluidVoice loads. Did you happen to take a look at that? 

In the current update I added a waiter so it waits to make sure the update happens clearly so - from next update it should fix this better

sfasano:
Thanks for taking a look! 

RE: memory: yes, even if the process is forcefully killed, the macOS kernel will completely reclaim all unified memory and MLX buffers assigned to that PID. It won't cause a system-level memory leak. 

However, with the latest commits on this PR, the `exit(0)` acts strictly as a 'dead man's switch' fallback. We call `NSApp.terminate(nil)` first, which triggers `AppDelegate.applicationWillTerminate` so the models gracefully unload themselves 99% of the time. The 20-second fallback timer only fires if the graceful cleanup completely deadlocks and refuses to quit.

That being said, if you've already implemented a proper waiter in the upcoming release that handles the update lifecycle more cleanly, feel free to just close this PR! I mainly wanted to get a safety net in place because the duplicate background instances were causing global hotkey collisions on my machine.

altic-dev:
Thanks a lot @sfasano  - let's try my current method to see if it fixes it - if not we'll come work on this together :) 

PS : we're also workin on a new updater to solve all this. That will solve all of the current issues. 

Really appreciate your support in this :)

PR #809, opened 2026-08-07
fix(typing): short-circuit verification for targets with no verifiable text

### LINKED ISSUE #802: [BUG] Consecutive dictations into Ghostty stall for seconds — pasteboard semaphore held across a 5 s verification that can never succeed in a terminal
### Describe the bug

When dictating repeatedly into Ghostty, the **first** insertion is instant (~45 ms). A **second** dictation started shortly after is transcribed correctly, but the text insertion stalls for **seconds** before the text appears.

The stall happens entirely inside the insertion call. Nothing is logged during it. The same rapid-fire pattern in a normal Cocoa text view (CotEditor) never stalls.

### Steps to Reproduce

1. Focus Ghostty (`com.mitchellh.ghostty`).
2. Dictate a short sentence, stop. Text appears immediately.
3. Within a few seconds, dictate a second short sentence and stop.
4. Observe the delay before the second text is inserted.
5. Repeat in CotEditor (or any standard `NSTextView`) — no delay.

### Actual behavior

Measured from my own `~/Library/Logs/Fluid/Fluid.log`, one session, 40 insertions:

- 37 insertions: `totalMs` between 42 and 53
- 3 insertions: `totalMs` = **2342**, **7450**, **8889**

All three slow insertions targeted Ghostty. All 8 CotEditor insertions in the same session completed at `totalMs=0`.

The full delay sits between `insert_call` and `insert_return`, with no log output in between:

```
[14:53:10.483] [INFO] [AppBenchmark]    APP_BENCH    text_ready chars=31
[14:53:10.483] [DEBUG] [ContentView]    Typing decision → frontmost: Ghostty, fluidFrontmost: false, editorFocused: true, willTypeExternally: true
[14:53:10.483] [INFO] [AppBenchmark]    APP_BENCH    focus_restore_result activated=false element=true elapsedMs=0 reason=already_focused
[14:53:10.483] [INFO] [TypingBenchmark] TYPING_BENCH request chars=31 mode=standard autocompleteSteps=1 preferredPID=2181 textReadyAgeMs=8
[14:53:10.483] [INFO] [TypingBenchmark] TYPING_BENCH worker_start queueDelayMs=0
[14:53:10.483] [INFO] [TypingBenchmark] TYPING_BENCH settle_delay_done delayMs=0 elapsedMs=0
[14:53:10.483] [INFO] [TypingBenchmark] TYPING_BENCH insert_call
[14:53:19.372] [INFO] [TypingBenchmark] TYPING_BENCH insert_return elapsedMs=8889 totalMs=8889
```

ASR is not involved: `final_done elapsedMs=48`, `text_ready` 8 ms later, focus restore 0 ms, queue delay 0 ms.

### Expected behavior

Consecutive dictations into a terminal insert as fast as into any other app. Clipboard restoration is an internal concern and must not block the next insertion.

### Root cause

In `Sources/Fluid/Services/TypingService.swift`:

1. Ghostty is detected by bundle identifier and, in `standard` mode, forced onto the reliable-paste path:

   ```swift
   if self.textInsertionMode == .standard,
      let ghosttyTargetPID = self.ghosttyTargetPID(preferredTargetPID: preferredTargetPID)
   {
       self.log("[TypingService] Ghostty target detected in standard mode (PID \(ghosttyTargetPID)); forcing Reliable Paste path")
   ```

2. That path calls `withTemporaryPasteboardString(text, restoreDelayMicros: 5_000_000)`.

3. `withTemporaryPasteboardString` acquires the process-global `pasteboardSessionSemaphore` (`DispatchSemaphore(value: 1)`) on entry and deliberately does **not** release it on return — it sets `releasesPasteboardSessionOnReturn = false` and hands the release to an async task on `pasteboardRestoreQueue`.

4. That task first runs `waitForFocusedTextVerification(..., timeoutMicros: 5_000_000)`, polling every 50 ms, before signalling the semaphore.

5. `waitForFocusedTextVerification` can only succeed via one of four checks — `appScriptValue.contains(expectedText)`, an AppleScript caret-distance match, `current.value.contains(expectedText)`, or an AX caret-distance match. **A GPU-rendered terminal exposes neither an AX text value nor a text-field-style selected range**, so none of the four can ever match. The loop always runs the full 5 s and returns `.timeout`.

6. The next insertion blocks at `pasteboardSessionSemaphore.wait()` for the remainder of that window. With several dictations in quick succession the windows stack, which is where the 7–9 s figures come from.

So the stall is not Ghostty being slow — it is a global lock held for a fixed 5 s whenever the target app cannot satisfy the verification, which is *always* true for terminals.

### Suggested fix

The semaphore currently guards two unrelated concerns: (a) not clobbering another session's pasteboard snapshot, and (b) waiting long enough before restoring the clipboard. Only (a) needs to serialize insertions, and it does not need to be held across the poll.

Options, roughly in order of preference:

1. **Release the semaphore once the paste has been dispatched and the snapshot is owned**, and let the restore task carry the snapshot as its own state. Serialize restores on `pasteboardRestoreQueue` rather than blocking the insertion path.
2. **Skip verification entirely when the target cannot support it** — if `captureFocusedTextSnapshot()` yields no usable AX text value and no selected range (the terminal case), fall back to a short fixed delay instead of a 5 s poll. Ghostty is already special-cased by bundle ID, so the information is available at that point.
3. **Bound the wait for a queued insertion**: if `pasteboardSessionSemaphore.wait(timeout:)` expires, proceed rather than block indefinitely.

Note that switching `TextInsertionMode` is not a workaround: `standard` explicitly routes Ghostty onto the reliable-paste path, and `reliablePaste` uses the same function with the same 5 s value.

### Environment

- FluidVoice: **1.6.7**
- macOS: **26.6 (25G72)**, Apple Silicon (arm64)
- Target app: **Ghostty 1.3.1** (`com.mitchellh.ghostty`)
- Model: Parakeet TDT v3 (Multilingual)
- `TextInsertionMode = standard` ("Clipboard Free Insert")

### Possibly related

- #479 — dictated list not inserted into Claude Code under cmux (same area: text insertion into terminals)
- #676 — overlay hides before insertion completes (cosmetic, does not explain the multi-second stall)

YuriNachos:
## Description

On targets that expose no verifiable accessibility text — GPU-rendered terminals such as Ghostty — `waitForFocusedTextVerification` polled for the full 5 s `timeoutMicros` while holding the process-global pasteboard session semaphore. That is the multi-second stall measured in #802 (sub-second → ~8.9 s) whenever dictation was injected into such a target.

The fix short-circuits verification when none of the four verifiable focused-text channels is present (`value`, `selectedRange`, `appScriptValue`, `appScriptSelectedRange` all `nil`): sleep a pasteboard-consume window, then release the session as `.unavailable`. Verifiable targets — any one channel present — keep their existing fast path unchanged.

The consume window is **500 ms** (`pasteConsumeWindowMicros`). `action()` only *enqueues* Cmd+V via `postToPid`; the target then reads `NSPasteboard.general` asynchronously on its own main thread, and under load (a large paste, a busy main thread during a build) that read can land hundreds of milliseconds later — so restoring the prior pasteboard too eagerly would clobber the transcript before the target consumes it. 500 ms is a conservative floor for a busy terminal to service the paste, yet ~10× under the 5 s timeout, so the #802 stall is gone.

Diff: `Sources/Fluid/Services/TypingService.swift` plus one test file.

## Type of Change

- [x] 🐞 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 🧹 Chore
- [ ] 📝 Documentation update

Closes #802.

## Testing

- [ ] Tested on Intel Mac
- [x] Tested on macOS version: 26.0 (Darwin 25.6.0), Xcode latest-stable
- [x] Ran linter locally: `swiftlint --strict --config .swiftlint.yml Sources` → **0 violations, 0 serious in 139 files**
- [x] Ran tests locally: `xcodebuild test -project Fluid.xcodeproj -scheme Fluid -destination 'platform=macOS,arch=arm64' CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO -skip-testing:FluidDictationIntegrationTests/DictationE2ETests/testDictationEndToEnd_whisperTiny_transcribesFixture`

Two new cases, both in `TypingServiceTransientPasteboardTests` (suite passes in full):

- `testFocusedTextSnapshotWithoutVerifiableChannelShortCircuitsVerification` — pins the predicate: all-four-nil is unverifiable; any single non-nil channel is verifiable.
- `testWaitForFocusedTextVerificationShortCircuitsUnverifiableTargetInBoundedTime` — **behavioural**: drives the real `waitForFocusedTextVerification` with an unverifiable snapshot (real pid, all four channels nil) and asserts it returns `.unavailable` in under `timeoutMicros / 2`. Remove the short-circuit and it polls to `.timeout` at ~5 s, failing both asserts.

Two unrelated tests are red on my machine only, for the environment reasons this repo's own prerequisites document, and both are red on a clean `main` here too:

- `HotkeyShortcutTests.testKeyboardPayloadIgnoresStrayMouseButtonField` — asserts the ASCII glyph at a keyCode; my host has a non-U.S. input source active.
- `DictationE2ETests.testAppPromptBinding_defaultFallbackIgnoresGlobalSelection` — reads real `com.FluidApp.app` defaults, which are customised on this machine.

Neither touches the typing-verification path. CI runs on a clean U.S.-layout runner, so both should be green there.

## Screenshots / Video

This is a backend typing-verification timing path — no view, layout, styling, settings, overlay or menu-bar surface is touched. The user-observable effect is latency: dictating into Ghostty returns in well under a second instead of ~8.9 s.

## Notes

- Verifiable targets are byte-identical to `main`; only the all-channels-nil case takes the new branch.
- The consume window is a named constant (`pasteConsumeWindowMicros`) so it is easy to tune if a slower target turns up.
- A few `private` declarations (`FocusedTextSnapshot`, `PasteVerificationResult`, `waitForFocusedTextVerification`) became `internal` so the short-circuit path can be exercised by tests; none of them are public API.

### DIFF
--- a/Sources/Fluid/Services/TypingService.swift
+++ b/Sources/Fluid/Services/TypingService.swift
@@ -1107,6 +1107,26 @@ final class TypingService {
         let selectedRange: CFRange?
     }
 
+    /// Returns `true` iff at least one of the four verifiable focused-text channels is present.
+    ///
+    /// A snapshot with a real `pid` but every verifiable channel `nil` — e.g. a GPU-rendered
+    /// terminal such as Ghostty that exposes neither an accessibility text value nor a text-field
+    /// selected range — cannot be verified, so `waitForFocusedTextVerification` short-circuits to
+    /// `.unavailable` instead of polling the full `timeoutMicros` while holding the pasteboard
+    /// session semaphore. The four channels are passed as primitive optionals so the private
+    /// `FocusedTextSnapshot` struct stays private.
+    static func focusedTextSnapshotSupportsVerification(
+        value: String?,
+        selectedRange: CFRange?,
+        appScriptValue: String?,
+        appScriptSelectedRange: CFRange?
+    ) -> Bool {
+        value != nil
+            || selectedRange != nil
+            || appScriptValue != nil
+            || appScriptSelectedRange != nil
+    }
+
     private func waitForFocusedTextVerification(
         from snapshot: FocusedTextSnapshot?,
         expectedText: String,
@@ -1114,6 +1134,20 @@ final class TypingService {
     ) -> PasteVerificationResult {
         guard let snapshot else {
             usleep(timeoutMicros)
+            return .unavailable
+        }
+
+        // Unverifiable target (e.g. a GPU-rendered terminal): no verifiable AX text value and no
+        // text-field selected range. The paste was already dispatched synchronously by `action()`
+        // before this restore task runs, so give it a single poll interval to be read, then release
+        // the pasteboard session without polling the full `timeoutMicros`.
+        if !Self.focusedTextSnapshotSupportsVerification(
+            value: snapshot.value,
+            selectedRange: snapshot.selectedRange,
+            appScriptValue: snapshot.appScriptValue,
+            appScriptSelectedRange: snapshot.appScriptSelectedRange
+        ) {
+            usleep(min(timeoutMicros, 50_000))
             return .unavailable
         }
 
--- a/Sources/Fluid/Services/TypingService.swift
+++ b/Sources/Fluid/Services/TypingService.swift
@@ -32,7 +32,12 @@ final class TypingService {
         let items: [PasteboardItemSnapshot]
     }
 
-    private struct FocusedTextSnapshot {
+    /// Captured before a paste so verification can diff the focused target against it.
+    ///
+    /// Internal (not private) only so behavioral tests can construct an unverifiable snapshot — a
+    /// real `pid` with all four verifiable channels `nil`, the terminal case — to exercise the
+    /// short-circuit in `waitForFocusedTextVerification`. It is not part of any public contract.
+    struct FocusedTextSnapshot {
         let pid: pid_t
         let bundleIdentifier: String?
         let value: String?
@@ -41,7 +46,9 @@ final class TypingService {
         let appScriptSelectedRange: CFRange?
     }
 
-    private enum PasteVerificationResult: String {
+    /// Internal (not private) only so behavioral tests can assert the short-circuit result
+    /// (`.unavailable`). It is not part of any public contract.
+    enum PasteVerificationResult: String {
         case appScriptContainsText = "appscript_contains_text"
         case appScriptCaretMovedExpectedDistance = "appscript_caret_moved_expected_distance"
         case fieldContainsText = "field_contains_text"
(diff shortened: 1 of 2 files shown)


YuriNachos:
Saw the 1.6.9 fixes ship earlier today — nice to see the release come together. This PR has been green for most of a week now (build, SwiftLint, the policy check and Greptile all pass), so I wanted to ask whether anything in the approach gives you pause — for example the 500 ms paste-consume window before the clipboard is restored — or whether it just hasn't had a review pass yet. Happy to make the window tunable, extend the behavioural tests, or rebase onto the release if any of that would help.

PR #815, opened 2026-08-08
fix(hotkey): stop modifier-only shortcut from firing on unrelated Shift+key combos

### LINKED ISSUE #688: [BUG] Left Option modifier-only hotkey falsely triggers recording on Shift+Enter
### Describe the bug

**Summary**
I set the primary dictation hotkey to **left Option** (modifier-only). Recording starts as expected when I tap left Option alone. However, pressing **Shift+Enter** — which I use constantly to insert a newline in text fields — also starts recording. Shift+Enter does not involve the Option key at all, so this looks like a false positive in the modifier-only shortcut matcher.

**Suspected cause (for triage)**
Because a single modifier can't be registered as a normal hotkey, the CGEvent tap in `GlobalHotkeyManager.swift` has to detect the left-Option hotkey from the raw `flagsChanged` / keyDown / keyUp stream (tap-detection). It looks like the interleaved Shift `flagsChanged` events around the Return keypress are being misread as an isolated modifier tap matching the configured Option shortcut. This is the same area touched by the modifier-only fixes in v1.5.12 (fn+control) and the right-side modifier fixes in v1.6.2 (Right Option + Right Shift).

Happy to test a beta build or provide logs if useful.

### Reproduction steps

1. In Settings, assign the primary dictation shortcut to **left Option** (single modifier).
2. Place the cursor in any multi-line text field.
3. Press **Shift+Enter** to insert a line break.
4. Observe that FluidVoice starts recording.

### Expected behavior

Shift+Enter inserts a newline and does not trigger dictation. Only a genuine left-Option tap (or hold) should trigger it.

### Actual behavior

Shift+Enter triggers dictation as if left Option had been tapped, interrupting normal typing.

### App Version

1.6.5 (16)

### macOS Version

MacOS Tahoe 26.5.2

### Architecture

Apple Silicon

### Logs or crash report

```text

```

_No response_

YuriNachos:
## Description
When the primary dictation hotkey is a single modifier (e.g. Left Option), an unrelated
Shift+key combo (Shift+Enter, Shift+R) could falsely start recording.

Root cause: the modifier-only start path in `GlobalHotkeyManager` was re-entrant. While the
configured modifier was held together with an extra modifier (Shift), releasing the extra
modifier shrank the pressed-modifier set back to just the configured modifier, which re-triggered
the start block and reset the "another key was pressed" flag. The subsequent release of the
configured modifier then looked like a clean tap and started recording.

The fix gates the start on the `flagsChanged` actually being for the configured modifier, so
releasing an unrelated modifier can no longer re-arm the press or erase the interrupt flag. A
genuine tap (and multi-modifier chords) still work. The start/finish decision was extracted into a
pure, unit-testable helper so a regression test could be added without instantiating the full
global event-tap manager.

## Type of Change
- [x] 🐞 Bug fix

Fixes #688

## Testing
- [ ] Tested on Intel Mac
- [x] Tested on macOS version: macOS 26.6 (Tahoe, Darwin 25.6.0)
- [x] Ran linter locally: `swiftlint --strict --config .swiftlint.yml Sources` → 0 violations in 139 files
- [ ] Ran formatter locally: `swiftformat` is not a CI check for this repo (swiftlint --strict is the authoritative gate)
- [x] Ran tests locally: `xcodebuild test … -only-testing:FluidDictationIntegrationTests/HotkeyShortcutTests` → 43 passed, 0 failures (under a U.S. input source)

## Screenshots / Video

## Notes
Pure logic fix in the modifier-only state machine — no UI, no hardware, no recording-pipeline
change. `HotkeyHoldModeType` was made internal and the start/finish decision was extracted into
`ModifierOnlyShortcutFlagsDecision.evaluate` to make the state machine unit-testable (the manager
can't be constructed in a test). Behavior-equivalence of the extraction was verified against the
original method; the only intentional behavior change is the new start guard.

### DIFF
--- a/Sources/Fluid/Services/GlobalHotkeyManager.swift
+++ b/Sources/Fluid/Services/GlobalHotkeyManager.swift
@@ -1,7 +1,7 @@
 import AppKit
 import Foundation
 
-private nonisolated enum HotkeyHoldModeType: Hashable {
+nonisolated enum HotkeyHoldModeType: Hashable {
     case transcription
     case promptMode
     case commandMode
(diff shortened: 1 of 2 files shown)


altic-dev:
oh wow - were you able to reproduce this or is it just a guess and a fix? Thanks for the PR!

YuriNachos:
Yes — reproduced it before writing the fix. The regression test in `Tests/FluidDictationIntegrationTests/HotkeyShortcutTests.swift` covers the exact path: with the primary hotkey set to a single modifier, holding Shift+key re-armed the start block because the modifier-only start path was re-entrant (releasing the extra modifier shrank the pressed-modifier set back to just the configured one). That test fails on the old code and passes after the fix.

PR #830, opened 2026-08-11
fix: preserve streamed tool calls with reasoning content

### LINKED ISSUE #445: [🐞 BUG] Command Mode fails with reasoning models when streamed tool calls are dropped or reasoning settings are mismapped
Describe the bug
  Command Mode is unreliable with some local reasoning-capable models. In my setup, models like qwen3.5:9b and gemma4:12b often fail in a way that looks silent: the model starts reasoning, but the final tool call either never gets executed or gets lost before FluidVoice can use it.

  I also found that the reasoning settings in the UI do not behave consistently across providers/models. On my local OpenAI-compatible server, reasoning_effort = none reliably changes the stream into a simpler, working format, but enable_thinking = false does not seem to have the same effect. Some values also appear to be available in FluidVoice but are rejected by the backend.

  To Reproduce

  1. Use a local OpenAI-compatible provider.
  2. Select a reasoning-capable model such as qwen3.5:9b or gemma4:12b.
  3. Enable Command Mode.
  4. Try a task that should trigger a tool call, such as checking files or running a terminal command.
  5. Leave AI streaming enabled.
  6. Observe that the model may stream reasoning first and then fail to complete the command reliably.
  7. Change the reasoning settings between reasoning_effort and enable_thinking.
  8. Observe that the behavior is not consistent across settings.

  Expected behavior
  FluidVoice should either:

  - preserve streamed tool calls even when the model emits reasoning first, or
  - map reasoning settings consistently so the selected model behaves correctly and predictably.

  Screenshots
  Optional.

  Environment:

  - macOS Version: Tahoe 26.5
  - App Version: 1.6.0
  - Architecture: Apple Silicon

  Additional context
  On my setup, reasoning_effort = none is the only setting that reliably makes some models behave in a Granite-like way. enable_thinking did not reliably change the stream behavior. In logs, the raw stream sometimes contains a valid tool call, but Command Mode still ends up with empty content or no usable tool call, which suggests a
  streaming/parser issue in FluidVoice rather than only a model quality issue.

  Crash Logs
  No crash. The app stays alive, but Command Mode fails or appears stuck.

postoso:
## Description
Some chat-completions providers stream reasoning content, an empty content string, and tool calls in the same delta. The parser advanced to the next SSE event after handling content, so the sibling tool call was dropped. This keeps the reasoning/content paths mutually exclusive while still parsing tool calls from the same delta.

## Type of Change
- [x] 🐞 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 🧹 Chore
- [ ] 📝 Documentation update

Fixes #445.

## Testing
- [ ] Tested on Intel Mac
- [x] Tested on macOS version: 15.7.7
- [ ] Ran formatter locally: `swiftformat --config .swiftformat Sources` (SwiftFormat is not installed locally)
- [x] Ran tests locally: `xcodebuild test -project Fluid.xcodeproj -scheme Fluid -destination 'platform=macOS,arch=arm64' -only-testing:FluidDictationIntegrationTests/LLMClientStreamingTests CODE_SIGN_IDENTITY='' CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO`

The two focused regression tests pass. Reintroducing the skipped-event behavior makes the same-delta tool-call test fail, and restoring this patch makes both tests pass again.

## Screenshots / Video

## Notes
This intentionally addresses the streamed tool-call loss without changing provider-specific reasoning settings.

### DIFF
--- a/Sources/Fluid/Services/LLMClient.swift
+++ b/Sources/Fluid/Services/LLMClient.swift
@@ -75,6 +75,11 @@ final class LLMClient {
         self.session = URLSession(configuration: config)
     }
 
+    /// Test seam for deterministic transport fixtures; callers own the session configuration.
+    init(session: URLSession) {
+        self.session = session
+    }
+
     // MARK: - Response Types
 
     struct Response {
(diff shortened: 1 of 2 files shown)


altic-dev:
merging without checking trusting you! hope this fixes the issue that's attached!!

Issue #219, opened 2026-03-22
[✨ FEATURE] Please provide ability to transcribe multiple voice memos all at once (voice memos.app)


fad1:
Recording with apple watch, which syncs voice memos to mac, it's possible to drag and drop memos as audio files one by one into finder, then into fluidvoice. It'd be great to be able to process them all at once. I used AI to describe the issue after it explored the FluidVoice codebase:

---

### Problem

Currently, transcribing Voice Memos recordings in FluidVoice requires a two-step workaround:

1. Drag a recording from Voice Memos onto Finder (to materialize the file) one by one
2. Drag the file from Finder into FluidVoice's File Transcription window

Dragging directly from Voice Memos to FluidVoice does nothing — the drop is silently ignored. The Voice Memos right-click Services menu also has no FluidVoice option.

This matters because Voice Memos' built-in transcription doesn't handle non-English languages well, while FluidVoice (via Parakeet/Whisper) handles multilingual audio much better.

### Root cause

Voice Memos uses **file promises** (`NSFilePromiseProvider`) for drag operations rather than putting a real file URL on the pasteboard. FluidVoice's drop handler only accepts `UTType.fileURL`, so it never sees the promised file. Finder resolves the promise into an actual `.m4a` file, which is why the two-step workaround works.

Relevant code: `MeetingTranscriptionView.swift` line 217:
```swift
.onDrop(of: [.fileURL], isTargeted: self.$isDropTargeted) { providers in
```

#### 1. Support file promises in drag & drop

Add `NSFilePromiseReceiver` handling to the existing drop target. This resolves the promise into a temporary file, then feeds it to the transcription pipeline — fixing direct drag from Voice Memos (and any other app that uses file promises, like Photos, Mail attachments, etc.).

This is the minimal fix for the direct drag & drop issue.

#### 2. Enable batch file selection

The file picker currently sets `allowsMultipleSelection: false`. Changing this to `true` and handling the array of URLs would allow navigating to Voice Memos' storage directory and selecting multiple recordings at once for sequential transcription.

Voice Memos stores recordings at:
```
~/Library/Group Containers/group.com.apple.VoiceMemos.shared/Recordings/
```

#### 3. "Import from Voice Memos" button (stretch goal)

[...]

PR #835, opened 2026-08-11
fix(diarization): stop overlap merge from shrinking speaker turns

YuriNachos:
## Description
`SpeakerDiarizationService.mergeAdjacentTurns` welded consecutive same-speaker turns by taking the incoming turn's `endSeconds` unconditionally. Diarization segments can overlap at boundaries (overlapping speech, VAD imprecision), so a later same-speaker turn that ends *inside* the accumulated turn overwrote the end with an earlier timestamp and silently dropped the trailing audio range `[turn.end, current.end]` from the transcribed slice. The fix takes `max(current.endSeconds, turn.endSeconds)`, so a merge can never reduce the covered time range.

## Type of Change
- [x] 🐞 Bug fix

Hardens the speaker-labeled file-transcription path landed in #378 (originally requested in #219). No standalone issue exists for the overlap case, so linking the feature it belongs to.

## Testing
- [ ] Tested on Intel Mac
- [x] Tested on macOS version: macOS 26 (Tahoe), arm64
- [x] Ran linter locally: `swiftlint --strict --config .swiftlint.yml Sources` → 0 violations on the edited files
- [ ] Ran formatter locally: `swiftformat --config .swiftformat Sources` (not a CI check; skipped to avoid unrelated formatting churn)
- [x] Ran tests locally: `xcodebuild test -project Fluid.xcodeproj -scheme Fluid -destination 'platform=macOS,arch=arm64' -only-testing:FluidDictationIntegrationTests/SpeakerTurnMergingTests` → 10 tests, 0 failures. The two new cases (`testOverlappingSameSpeakerTurnDoesNotShrinkMergedRange`, `testTouchingSameSpeakerTurnExtendsToEnd`) fail on the previous `endSeconds: turn.endSeconds` and pass with `max(...)`.

## Screenshots / Video

## Notes
Pure logic change to a `static`, already unit-tested merge helper — no SettingsStore, UI, or model loading involved. Behavior for the normal non-overlapping case is unchanged (a later turn's end is `>=` the accumulated end there, so `max` collapses to the previous value).

### DIFF
--- a/Sources/Fluid/Services/SpeakerDiarizationService.swift
+++ b/Sources/Fluid/Services/SpeakerDiarizationService.swift
@@ -99,10 +99,15 @@ actor SpeakerDiarizationService {
                gap <= maxGapSeconds,
                combinedDuration <= maxTurnSeconds
             {
+                // A merge must never shrink the covered time range. Diarization
+                // segments can overlap at boundaries (overlapping speech, VAD
+                // imprecision), so a later same-speaker turn may end before the
+                // accumulated turn — taking turn.endSeconds directly would drop the
+                // trailing audio ([turn.end, current.end]) from the slice.
                 current = SpeakerTurn(
                     speakerLabel: current.speakerLabel,
                     startSeconds: current.startSeconds,
-                    endSeconds: turn.endSeconds
+                    endSeconds: max(current.endSeconds, turn.endSeconds)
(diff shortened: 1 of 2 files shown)


altic-dev:
@YuriNachos  you're awesome!

PR #860, opened 2026-08-14
Batch A: session-history data-loss fix + first-run landmines

stalinzbb:
First batch from the post-v1 audit: the data-loss bug plus every first-five-minutes landmine, and a test-infrastructure fix the suite needed to run at all on a machine with FluidVoice installed.

## Data loss (P0-1)
- `PracticeSessionStore.load` wiped the **entire** session history when any one entry failed to decode; the non-optional `quality` field added with the signal-quality gate already breaks every pre-gate file.
- Entries now decode individually (a corrupt entry drops that entry only), and `DeliveryMetrics` decodes a missing `quality` as `.unknown`, so pre-gate sessions keep their measurements, flagged as never-assessed.
- Both behaviors pinned by new tests.

## First-run landmines
- **Onboarding no longer hard-blocks on Accessibility** — practice never types into other apps; the permissions step gates on microphone only, AX rows are labelled "(Optional)" with an honest explanation, and the hotkey playground step no longer blocks Continue. (P1-1)
- **Option+R rewrite hotkey now defaults off** — a fresh install no longer owns a global AX text-rewrite chord in every app. (P1-2)
- **macOS permission dialogs no longer say "FluidVoice"** — all four usage strings rewritten for Voix. (P1-3)
- **Deleted `stale.yml`** (auto-closed stacked PRs into the unrecoverable state BRANCHING.md warns about) and **`FUNDING.yml`** (routed sponsorship to upstream); the two in-app support blurbs now point at this fork's issues instead of upstream's email. (P1-10)

## Test infrastructure
The suite was failing with "test runner hung before establishing connection" — the unsigned test host blocked forever on a keychain authorization prompt, because `KeychainService` still uses upstream's service name and this machine now has a real key under it. Under XCTest the keychain primitives now use an in-memory store. The shared service name itself is a real product bug (Voix and installed FluidVoice clobber each other's keys) — tracked as a separate task.

## Verification
- 174/174 tests pass locally (E2E suite included), where the suite previously could not run at all.
- Manual checks per the plan: fresh onboarding completes with mic permission only; pre-gate session file loads with sessions intact.


stalinzbb:
Apologies — opened against the wrong repo (this belongs on a fork). Closing.

PR #862, opened 2026-08-14
fix(command-mode): close three confirm-gate bypasses: absolute path, find -delete/-exec rm, diskutil

### LINKED ISSUE #861: isDestructiveCommand's confirm gate has three bypass classes PR #434 doesn't cover: absolute-path invocation, find -delete/-exec rm, and diskutil
## Description

`CommandModeService.isDestructiveCommand` (Sources/Fluid/Services/CommandModeService.swift:562, current main) decides whether an AI-issued shell command in Command Mode requires user confirmation before running. Three ways a plausible (not adversarial-looking) model output skips it entirely:

**1. The primary command invoked via an absolute path.** The prefix list only matches bare names (`"sudo "`, `"mv "`, `"chmod "`, etc.). `/usr/bin/sudo reboot`, `/bin/mv secret.txt /tmp/`, `/bin/chmod 000 /etc/hosts`, `/usr/bin/killall Finder`, `/bin/rmdir ~/Documents` all skip the check their bare-name equivalents trigger. `rm` happens to have a partial fallback (`cmd.contains("rm -")`) that catches most `/bin/rm -rf` cases, but not `/bin/rm somefile` (no dash flag), and no other command has an equivalent fallback at all.

**2. `find -delete` / `find ... -exec rm ...`.** Never matches any prefix or the `|`/`;`/`&&` patterns, since `rm` inside `-exec` never sits next to a matched separator. `find ~/Documents -delete` skips detection with no path trick needed at all.

**3. `diskutil`'s destructive subcommands.** `dd `, `mkfs`, `format` are the only disk-operation prefixes. `diskutil eraseDisk`, `diskutil secureErase`, `diskutil eraseVolume` are a completely different binary and aren't covered anywhere.

## Relationship to #434

#434 ("flag pipe-to-shell, output-redirect, and whitespace-prefixed destructive commands in the confirm gate") is open and does extensive work on this same function, but on the pipe-target side of the classification, not the leading-command matching logic. I checked this rather than assumed it: fetched #434's actual branch, mechanically extracted its patched `isDestructiveCommand` plus every helper it calls (its own diff moves the function from `private func` to `nonisolated static func` for testability, which is also what I did independently before checking #434 had already made the same call), and reran the same adversarial inputs against #434's real code:

```
$ swift pr434_wrapped.swift
=== against PR #434's actual patched code: absolute-path bypass ===
caught("sudo reboot")=true   bypass("/usr/bin/sudo reboot")=false   STILL BYPASSED
caught("mv secret.txt /tmp/")=true   bypass("/bin/mv secret.txt /tmp/")=false   STILL BYPASSED
caught("chmod 000 /etc/hosts")=true   bypass("/bin/chmod 000 /etc/hosts")=false   STILL BYPASSED
caught("killall Finder")=true   bypass("/usr/bin/killall Finder")=false   STILL BYPASSED
caught("rmdir ~/Documents")=true   bypass("/bin/rmdir ~/Documents")=false   STILL BYPASSED

=== against PR #434's actual patched code: find -delete/-exec ===
flagged=false   cmd="find ~/Documents -delete"
flagged=false   cmd="find ~/Documents -type f -delete"
flagged=false   cmd="find / -name '*.important' -exec rm {} \;"

=== against PR #434's actual patched code: diskutil ===
flagged=false   cmd="diskutil eraseDisk JHFS+ Untitled disk0"
flagged=false   cmd="diskutil secureErase 0 /dev/disk0"
```

All three classes survive #434's hardening unchanged, so this isn't overlapping work — it's a genuinely separate gap in the same function. I'm filing this as an issue rather than a competing PR against #434 to avoid a merge conflict on the same lines; happy to have the patch below folded into #434 or landed separately, whichever you'd prefer.

## Proposed fix

Resolves the leading command to its bare name the same way a shell would (last path component of the first token), so `/bin/rm`, `/usr/bin/rm`, and bare `rm` are all recognized as the same command regardless of how the model referenced it, plus dedicated checks for the `find` and `diskutil` cases. Scoped narrowly — `diskutil list`/`diskutil info` (read-only) are explicitly not flagged.

```diff
--- a/Sources/Fluid/Services/CommandModeService.swift
+++ b/Sources/Fluid/Services/CommandModeService.swift
@@ -437,7 +437,7 @@ final class CommandModeService: ObservableObject {
                 }

                 // Check if we need confirmation for destructive commands
-                if SettingsStore.shared.commandModeConfirmBeforeExecute, self.isDestructiveCommand(tc.command) {
+                if SettingsStore.shared.commandModeConfirmBeforeExecute, Self.isDestructiveCommand(tc.command) {
                     self.pendingCommand = PendingCommand(
                         id: tc.id,
                         command: tc.command,
@@ -559,7 +559,7 @@ final class CommandModeService: ObservableObject {
         }
     }

-    private func isDestructiveCommand(_ command: String) -> Bool {
+    nonisolated static func isDestructiveCommand(_ command: String) -> Bool {
         let cmd = command.lowercased()

         // Commands that start with these are destructive
@@ -598,6 +598,49 @@ final class CommandModeService: ObservableObject {
             return true
         }

+        // The prefix list above only matches a bare command name. A model
+        // that reaches for `/bin/rm`, `/usr/bin/sudo`, etc. (not unusual —
+        // absolute paths are a normal way to disambiguate a binary) skips
+        // every check above except the `rm -` fallback, which only happens
+        // to catch `rm` and only when it carries a `-` flag. Resolve the
+        // leading token to its bare command name the same way a shell would
+        // (last path component) so `/bin/rm`, `/usr/bin/rm`, and bare `rm`
+        // are all recognized as the same command regardless of how the
+        // model referenced it.
+        let leadingToken = cmd
+            .drop(while: { $0 == " " || $0 == "\t" })
+            .prefix(while: { $0 != " " && $0 != "\t" })
+        let commandName = (leadingToken as NSString).lastPathComponent
+        let destructiveCommandNames: Set = [
+            "rm", "rmdir", "mv", "sudo", "kill", "pkill", "killall",
+            "chmod", "chown", "chgrp", "dd", "mkfs", "shred", "truncate",
+        ]
+        if destructiveCommandNames.contains(commandName) {
+            return true
+        }
+
+        // `find -delete` / `find ... -exec rm ...` deletes without ever
+        // matching "rm -" or any `|`/`;`/`&&` pattern above, since `rm`
+        // inside `-exec` never sits next to a matched separator.
+        if commandName == "find", cmd.contains(" -delete") || (cmd.contains("-exec") && cmd.contains("rm ")) {
+            return true
+        }
+
+        // diskutil's erase/reformat/partition subcommands are as destructive
+        // as `dd`/`mkfs`/`format` but are a different binary entirely and
+        // weren't covered by any check above. Scoped to the destructive
+        // subcommands specifically so read-only uses (`diskutil list`,
+        // `diskutil info`) are not flagged.
+        if commandName == "diskutil" {
+            let destructiveDiskutilSubcommands = [
+                "erasedisk", "erasevolume", "secureerase",
+                "reformat", "partitiondisk", "zerodisk", "unmountdisk",
+            ]
+            if destructiveDiskutilSubcommands.contains(where: { cmd.contains($0) }) {
+                return true
+            }
+        }
+
         return false
     }
```

Plus a new test file, `Tests/FluidDictationIntegrationTests/CommandModeDestructiveCommandGapTests.swift`, covering: the three fixed bypass classes, a regression check that the original bare-command detection still works, and a false-positive check on benign commands (`diskutil list`, `find . -name '*.txt'`, `/usr/bin/python3 --version`, etc.) so the gate doesn't get more aggressive than it needs to be.

## Testing

- [ ] Tested on Intel Mac
- [x] Tested on macOS version: 26.1
- [x] Ran linter locally: `swiftlint lint --strict --config .swiftlint.yml Sources/Fluid/Services/CommandModeService.swift` — 0 violations. Same for the new test file.
- [x] Ran formatter locally: `swiftformat --config .swiftformat --lint` — clean on the changed region (the file has 37 pre-existing formatting violations elsewhere, none touched by this change, confirmed by running the same lint against the unmodified original file first)
- [x] Ran tests locally:

```
xcodebuild test -project Fluid.xcodeproj -scheme Fluid -destination 'platform=macOS' \
  -only-testing:FluidDictationIntegrationTests/CommandModeDestructiveCommandGapTests \
  CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO

Test Suite 'CommandModeDestructiveCommandGapTests' passed at 2026-08-14 12:34:45.767.
	 Executed 5 tests, with 0 failures (0 unexpected) in 0.006 (0.008) seconds
```

- [x] `xcodebuild -project Fluid.xcodeproj -scheme Fluid -configuration Debug build` on macOS 26.1 / Xcode 26.3 (Apple Silicon): **BUILD SUCCEEDED**, no new warnings.

## Notes

This is a defense-in-depth gap, not a way to get code execution FluidVoice doesn't already grant — Command Mode already gives its configured LLM real shell access by design, and the confirm gate exists to catch the model's own mistakes or a manipulated/hallucinated response before something destructive runs unattended. All three classes here let a plausible model output — using an absolute path, or `find`, or `diskutil`, none of which are unusual or adversarial-looking on their own — skip that one safety check silently.

manjunathbhaskar:
## Description

`isDestructiveCommand` decides whether an AI-issued shell command in Command Mode needs user confirmation before running. It was doing ad hoc string matching on the raw command — a prefix list of bare names, a hardcoded list of separator substrings like `"&& rm "`, and a `cmd.contains("rm -")` catch-all. That technique can't be made complete: every bypass found during review was a different instance of the same two root causes (only the first space-delimited token was ever examined, and quoting/escaping wasn't parsed at all).

This replaces it with an actual parse. Same file, same function, same signature, one call site — nothing else in the app changes.

## How it works now

1. Split the command into its constituent simple commands on unquoted `&&`, `||`, `;`, `&`, `|`, and newline, honoring quotes and backslash escapes.
2. Tokenize each one with real quote and escape tracking, so a quoted or escaped span stays a single word.
3. Resolve each command's `argv[0]` to its basename and classify **every** simple command in the chain, not just the leading token of the first.

Where a program can appear as an *argument* rather than as `argv[0]`, the same basename resolution applies — one shared helper used by `find -exec`/`-execdir`, `xargs`, `env`, `nohup`, `command`, `exec`, `time`, `nice`, `setsid`, `stdbuf`, `timeout`, and `watch`. A `sh`/`bash`/`zsh -c` payload is parsed recursively (depth-capped) rather than treated as an opaque string.

## What this closes

Bare names; absolute and relative path invocation; quoted paths, including with internal spaces; backslash-escaped paths; destructive commands anywhere in a compound chain, not just first; leading environment assignments (`LC_ALL=C rm -rf x`); redirects anywhere including as the entire command (`> file`); path-qualified programs handed to a runner (`xargs -0 /bin/rm`, `find . -exec /bin/rm {} \;`); nested shell payloads (`sh -c 'rm -rf x'`); `find -delete`/`-exec`; and `diskutil`'s destructive subcommands including behind the `quiet` modifier, without false-positiving on `diskutil info /Volumes/EraseDisk`.

## Known limits — deliberately out of scope

This is a blocklist over shell syntax, so it is bounded by what string analysis can see. It does **not** catch:

- **Non-shell interpreters.** `python3 -c "os.remove(...)"`, `osascript -e`, `perl -e`, `node -e`. The payload is a different language; reading it means statically analyzing arbitrary code. (Shell wrappers are handled because the payload is shell and this parser can read it.)
- **Variable and command substitution.** `X=rm; $X -rf ~`, `$(echo rm) -rf ~`. Catching these requires evaluating shell semantics, not parsing.
- **Symlink indirection.** `ln -s /bin/rm ~/t && ~/t -rf ~`. Needs runtime filesystem resolution.
- **Destructive commands not in the list.** `defaults delete`, `launchctl unload`, `tccutil reset`, `csrutil`, `nvram`. The name lists are hand-maintained and will always lag.

Closing those requires inverting the model — confirm everything except a short allowlist of known-safe commands — which is a product decision rather than a fix to this function, and out of scope here. This PR is a strict improvement over `main` within the stated boundary: everything it catches, `main` missed, and nothing previously caught was lost.

## Testing

23 tests in `Tests/FluidDictationIntegrationTests/CommandModeDestructiveCommandGapTests.swift`, covering every case above plus negative cases so benign commands don't start requiring confirmation. Each fix was verified fail-before/pass-after against the preceding commit.

The tokenizer was additionally fuzzed standalone outside XCTest — empty input, unterminated quotes, lone trailing backslashes, 100k-character strings, 10k repeated operators, null bytes, unicode and emoji, tab-only whitespace — no crashes and no slow paths (worst case 21 ms).

`swiftlint` clean. Full app build succeeds in Release configuration. The wider `FluidDictationIntegrationTests` target passes (138 tests) once past a pre-existing launch crash in `HotkeyShortcutTests` that reproduces on `main` and is unrelated to this change.

Closes #861


altic-dev:
please fix the issues and I can check if I Can merge it. THanks!

manjunathbhaskar:
fixed both review comments -- quoted absolute paths ("/bin/rm") now resolve correctly instead of keeping the trailing quote, and diskutil's subcommand check matches the actual subcommand token instead of searching the whole command string, so an argument like /Volumes/EraseDisk on a plain diskutil info call doesn't false-positive anymore. both covered by new tests, ran the full suite locally, everything green.

manjunathbhaskar:
fixed the quiet modifier gap too -- was only reading the first token after diskutil as the subcommand, so quiet eraseDisk classified quiet as the verb and let the erase through. now skips quiet before reading the subcommand. new tests cover both the quiet+destructive-verb case and quiet+read-only-verb staying unflagged.

manjunathbhaskar:
fixed all three from the latest round -- xargs target now resolves basename before comparing (was matching /bin/rm against bare rm and missing it), a bare redirect as the whole command (> file) is caught (redirect scan wasn't including the first word), and a leading env assignment (LC_ALL=C rm -rf x) now gets skipped before resolving the actual program instead of stopping at the assignment itself.

also: the two older codex comments (format/mkfs prefix, compound command after &&) are already covered by the tokenizer rewrite a few commits back -- format/mkfs.* is in the basename check now and every simple command in a chain gets checked, not just the first. tests for both exist (testFormatAndMkfsVariantBasenamesAreCaught, testDestructiveCommandAfterSeparatorIsCaught) and pass. greptile auto-resolved its own version of these, codex's threads just didn't follow.

re-fuzzed the tokenizer standalone after this round too -- degenerate assignment shapes, 40k+ char inputs, unicode, no crashes. full app build succeeds.

manjunathbhaskar:
@altic-dev fixed the latest three. worth calling out what the find -exec one actually was: it's the same bug as the xargs one from the previous round, three lines away, and i'd fixed one and not the other. so instead of patching the second instance i pulled it into one helper that every argument-position program reference now goes through -- find -exec/-execdir, xargs, env, nohup, command, exec, time, nice, setsid, stdbuf, timeout, watch. including ones nobody has flagged yet.

also fixed backslash-escaped paths in the tokenizer (/tmp/tools\ dir/rm), and sh -c payloads now get parsed recursively instead of treated as an opaque argument -- that one was a real regression i introduced, the old contains("rm -") caught it by accident and my rewrite didn't.

on the review loop generally: i've rewritten the PR description with an explicit "known limits" section, because i don't think these bots will ever return zero findings on a function like this and i'd rather the scope be stated than implied. non-shell interpreters (python3 -c, osascript -e), variable/command substitution (\ -rf), symlink indirection, and destructive commands not in the name list are all out of reach of string analysis and are documented as such. closing those properly means inverting the model to an allowlist, which is a product call rather than something to bolt onto this function.

within that boundary this is strictly better than main -- everything it catches main missed, nothing previously caught was lost, 23 tests each verified fail-before/pass-after, tokenizer fuzzed separately, full release build green. happy to keep going if you want any of the out-of-scope items pulled in, but i'd rather you make that call than keep expanding scope on my own.

altic-dev:
is this PR by claude specifically?

altic-dev:
There are like insane amount of comments and LOC for this work. down to take a look again if you wanna verify this! thanks!

manjunathbhaskar:
Yeah, AI assisted.
The original approach I tried to detect destructive patterns. And with each review round I found a new way a dangerous command could hide (absolute paths, quoting, xargs, find -exec/-delete, eval, nested shells, etc.), so the code kept growing.
Now I flipped the model instead: only a short, explicit allowlist of known-safe commands is to be trusted. Everything else requires confirmation by default. Redirects, command substitution, and shell metacharacters are also rejected up front.
The Result: smaller diff, fewer special cases, and the entire class of bypasses is closed structurally rather than one-by-one. The tradeoff is more confirmation prompts for unrecognized commands,  that’s intentional.

manjunathbhaskar:
Still around on this one if there's anything more to look at, just didn't want it to close on its own.

PR #864, opened 2026-08-15
docs: replace dead X handle link with GitHub org

yzxcj797:
The README links to `x.com/ALTIC_DEV` (badge and text) which returns 404 — the handle appears to have been removed. Updated both occurrences to the GitHub org (`github.com/altic-dev`), which is live and the natural place for project followers.


grohith327:
Can you change it to this instead https://x.com/fluidvoiceapp

Issue #891, opened 2026-08-24
[BUG] Left-click not working on opening the app


thegoodgamer14:
### Describe the bug

When trying to setup the shortcut to copy the last transcription on FluidVoice; I accidentally setup the shortcut as Cmd + Left-click. Since then, every time I open FluidVoice, the regular left-click on my Mac's trackpad stops working. This is a very peculiar bug because, I'm unable to do any click interactions within the app, so I'm not able to change the shortcut either. The left-click function is restored on quitting the app (Cmd + q). 

### Reproduction steps

1. Open FluidVoice
2. Go to Settings > Shortcuts > Copy Last Transcription
3. Set shortcut to "Cmd + Left-click" (at your own risk!)
4. Voila! you won't be able to use your trackpad when FluidVoice is open

### Expected behavior

Regular left-click should still work regardless of the shortcut set on the app.

### Actual behavior

Left-click stops working on opening the app. Functionality restored on quitting the app. Laptop becomes useless on opening FluidVoice.

### App Version

1.6.9

### macOS Version

macOS Tahoe 26.5.1

### Architecture

Apple Silicon

### Logs or crash report

```text

```

altic-dev:
this seems like an annoying issue - would you be down to clean up the saved data while I figure out an actual fix?

thegoodgamer14:
thanks for your response! Yes, I'm ok with that. How do I go about doing this? The app is not navigable by me on opening due to this issue. Do I uninstall and reinstall to clear the data (given it's stored locally)?

altic-dev:
@thegoodgamer14 

osascript -e 'tell application "FluidVoice" to quit'
defaults delete com.FluidApp.app PasteLastTranscriptionHotkeyShortcut
defaults write com.FluidApp.app PasteLastTranscriptionShortcutEnabled -bool false
open -a FluidVoice

try these out. it should clear that shortcut out while I try to fix the issue

altic-dev:
Hey, I’m tracing the mouse-shortcut and event-tap problems together now. Would you be willing to test a custom build on the Mac where this happened? I’d like to verify both the saved Cmd + Left-click case and normal clicking before shipping the fix.

Issue #884, opened 2026-08-21
[~BUG] Global trigger key still functions even when device is locked


samkelleher:
If you use global key bindings to trigger recording, these function as normal even if the device is locked . This has a slightly shady consequence of it recording and dictating whilst appearing to not be listening . The recording pill is not shown. 

### Reproduction steps

Lock your device and press your transcribe global key 

### Expected behavior

A locked device should not trigger any action 

### Actual behavior

FluidVoice is able to activate the microphone and start recording 🥷 

### App Version

1.6.9

### macOS Version

26.6

### Architecture

Apple Silicon

Issue #789, opened 2026-08-05
[BUG] Improper manifest key causes redundant and wasteful use of dock space


hillct:
### Describe the bug

Because FluidVoice is implemented as a service available from an icon and menu in the menu bar, the icon in the Dock is redundant and wasteful of screen real-estate. To resolve this and provide a more polished appearance to youur product the following manifest key should be set, which will have the result of hiding the Dock icon while retaining the menu bar item. It has the added effect of removing FluidVoice from the Cmd-Tab carousel which is debatably desirable, but in my estimation offers a much more polished look to the product. Set the manifest key:
```
<key>LSUIElement</key>
<true/>
```

### Reproduction steps

1. Launch FluidVoice
2. Observe the F icon in the menu bar, from which all the useful access points of the application are available
3. Observe the F icon with the black surround in the Dock, which is redundant, offering no real value beyond that offered by the menu bar item

### Expected behavior

The truly useful access point for the application should be the menu far icon (and submenu) as the Dock icon offers no uniquely useful interface, consumes dock real-estate and might even be confusing to users not familiar with Mac application architecture. Having the single useful acess point will also allow for simplification of documentation.

### Actual behavior

Upon launch, we see both the mac menu bar icon and submenu offering the full range of Fluidvoice feature access, and additionally, the redundant Dock icon which offers access only to the main screen (already accessible from the menu bar icon) and nothing else, except to waste dock real-estate and make the application appear un-polished.

### App Version

1.6.7 (18)

### macOS Version

26.5.2 (25F84)

### Architecture

Apple Silicon

### Logs or crash report

```text

```

hillct:
While I have not observed the workflow blocking behaviors described in https://github.com/altic-dev/FluidVoice/issues/753 that solution might conceivably address this concern as well. Still, there is no reason for the icon to even be optionally available in the dock, in my estimation.

hillct:
We do need to see some results here

hillct:
This issue was resolved in the previous release.

hillct:
Failing to resolve this issue in the proper manner suggests very poor product management. It seems that it might in fact have been resolved in the last major release, but the fact that you can't seem to track that issue correctly is a fundamental problem

altic-dev:
2 people working on this for free is not an easy job. we are taking care of things in the order of criticality. sometimes some issues might go unnoticed unfortunately and will get to it while we look deeper. 

[...]

Issue #657, opened 2026-07-19
[BUG] Global hotkey event tap becomes disabled and drops Function-key presses


kalenjordan:
### Describe the bug

The global hotkey `CGEventTap` can become disabled without the tap-disabled callback restoring it. While disabled, physical Function-key presses are not delivered to `GlobalHotkeyManager`, so an active recording cannot be stopped with the configured Function shortcut. The existing periodic health check eventually detects and recreates the tap, but its 30-second interval leaves the shortcut unresponsive for a noticeable period.

This was reproduced in both Automatic and Toggle activation modes.

### Reproduction steps

1. Configure Function as the primary dictation shortcut.
2. Start recording.
3. Continue using FluidVoice until the global event tap becomes disabled.
4. Press and release Function repeatedly while recording.
5. Observe that recording continues.
6. Wait for the health check to recreate the event tap.
7. Press Function again and observe that recording stops immediately.

The failure is intermittent, but occurred repeatedly in one session.

### Expected behavior

The Function shortcut remains available throughout recording. If macOS disables the event tap, FluidVoice detects and restores it immediately or within a very short bounded interval.

### Actual behavior

Multiple physical Function presses were absent from the application log while ASR remained active. The log then reported a failed event-tap health check, successfully recreated the tap, and the next Function press stopped recording.

### App Version

1.6.4 development build

### macOS Version

macOS 26.3 (25D125)

### Architecture

Apple Silicon

### Logs or crash report

```text
[20:42:14.057] [INFO] [ASRService] START() called - beginning recording session
# Multiple physical Function taps occurred here, with no GlobalHotkeyManager entries
[20:42:57.941] [WARN] [GlobalHotkeyManager] Health check failed, attempting to recover
[20:42:57.941] [INFO] [GlobalHotkeyManager] Event tap successfully created and enabled
[20:42:57.941] [INFO] [GlobalHotkeyManager] Health check recovery successful
[20:42:59.224] [INFO] [GlobalHotkeyManager] Hotkey route | pressed=dictate(mod) | active=dictate | asrRunning=true | action=stop
[20:42:59.226] [INFO] [ASRService] STOP() called - beginning shutdown sequence
```

Only one FluidVoice process was running. A local mitigation reduces the fallback health-check interval from 30 seconds to 2 seconds and adds detection/recovery timing diagnostics. A durable fix may require reducing work in the event-tap callback or safely moving event processing off the main run loop.

[...]

PR #934, opened 2026-09-02
Fix microphone, hotkey, and Dock reliability

### LINKED ISSUE #753: [BUG] Dock icon reappears on reopen despite "Hide from dock" — also blocks the hide-menu-bar-icon workflow (#382)
## Bug

With **"Hide from dock" enabled**, if FluidVoice is already running and I click its icon in Applications (or Launchpad/Spotlight), the Dock icon appears — and stays.

**To reproduce:**
1. Enable "Hide from dock" in Settings. Confirm no Dock icon.
2. With FluidVoice still running, open it again from the Applications folder.
3. The main window opens, and the Dock icon reappears permanently.

**Expected:** the main window opens; the Dock stays clean.

**Root cause (from reading the source):** a LaunchServices reopen resets the activation policy to the bundle default (`.regular`). `AppDelegate.swift` already handles this for the app's *own* programmatic reopen path (the comment in `requestMainWindowReopenIfNeeded()` describes exactly this, and re-applies `applyDockVisibilityPolicy()` afterward) — but an *external* reopen (user clicking the app icon) goes through `applicationShouldHandleReopen`, which never re-applies the policy.

**Suggested fix:** call `applyDockVisibilityPolicy()` in `applicationShouldHandleReopen` (or `applicationDidBecomeActive`).

## Why this matters — the workflow it unlocks (related: #382)

Like the users in #382, I'd love a settings toggle to hide the menu bar icon. Together with this bug fix, that enables the ideal minimal setup:

- FluidVoice runs with **no menu bar icon and no Dock icon** — fully invisible, just dictation via hotkeys.
- On the rare occasion I need settings, I open the app from Applications → the settings window appears, **still nothing in the Dock or menu bar**.

This reopen bug is a prerequisite: without the fix, reopening the app to reach settings permanently pollutes the Dock, which defeats the purpose of hiding the icons.

**Environment:** FluidVoice 1.6.6, macOS 26.5.2

### LINKED ISSUE #921: [BUG] Trackpad/mouse clicks stop registering system-wide minutes after login - filter-mode mouse event tap wedges (disabled-by-timeout at 311 s) with no mouse shortcut configured
### Describe the bug

FluidVoice installs two **filter-mode** (`kCGEventTapOptionDefault`) CGEventTaps whose event mask includes left mouse clicks, so every click system-wide is routed through FluidVoice before macOS delivers it. When the tap stops being serviced, trackpad and mouse clicks stop registering across the whole system - the cursor still tracks and the keyboard still works, but nothing can be clicked. Quitting FluidVoice restores clicks instantly.

On my machine this happens within minutes of every boot (FluidVoice is a login item), and I have **no mouse-button shortcut configured**, so clicks should arguably not be routed through the app at all.

I believe this ties several existing reports together:

- #891 - identical user-facing symptom (left-click dead while the app runs, restored on quit), reached via a saved Cmd+Left-click shortcut. Same delivery path, different trigger.
- #657 - the tap-disabled-by-timeout mechanism: the tap-disabled callback fails to restore the tap and only a 30-second health check eventually recreates it.
- #542 (mine, July) and #752 - the audio route-change deadlock family. My environment has frequent device-list churn (AirPods Pro, Studio Display sleep/wake, iPhone Continuity microphone appearing), which matches the trigger pattern.

### Evidence

Six minutes after login, `CGGetEventTapList` showed one of FluidVoice's two mouse-intercepting filter taps already **disabled by timeout after ~311 seconds of unprocessed event latency** (3.11e8 µs), i.e. the tap was wedged for essentially its whole life:

```
tap=783368690  pid=1062 proc=/Applications/FluidVoice.app/Contents/MacOS/FluidVoice
  enabled=true  options=FILTER (can block events) tapsMouseClicks=true
  minLatency=1416.2us avgLatency=1416.2us maxLatency=1416.2us
tap=1350490027 pid=1062 proc=/Applications/FluidVoice.app/Contents/MacOS/FluidVoice
  enabled=false options=FILTER (can block events) tapsMouseClicks=true
  minLatency=3.1138838e+08us avgLatency=3.1138838e+08us maxLatency=3.1138838e+08us
```

No other process on the system had a tap intercepting mouse clicks.

Notably, `sample` of the process at the same moment showed the **main thread idle** in the normal AppKit event loop (`mach_msg` under `-[NSApplication run]`) - so this is not the #542 main-thread-deadlock signature. The app looks alive; whatever run loop services the tap callback is starved, and (as in #657) the disabled-by-timeout callback does not recover it.

`~/Library/Logs/Fluid/Fluid.log` is silent for the entire wedge window - nothing between startup completing at 16:28:56 and 16:35:30, when device-list churn begins (`Device list changed`, iPhone Continuity microphone appearing in the input list).

Diagnostic script used (no root required), in case it is useful for reproducing:

```swift
import CoreGraphics
import Foundation

var count: UInt32 = 0
CGGetEventTapList(0, nil, &count)
var taps = [CGEventTapInformation](repeating: CGEventTapInformation(), count: Int(count))
CGGetEventTapList(count, &taps, &count)

let mouseMask = UInt64(1) << CGEventType.leftMouseDown.rawValue | UInt64(1) << CGEventType.leftMouseUp.rawValue
for t in taps {
    var name = "pid \(t.tappingProcess)"
    let path = UnsafeMutablePointer<CChar>.allocate(capacity: 4096)
    defer { path.deallocate() }
    if proc_pidpath(t.tappingProcess, path, 4096) > 0 { name = String(cString: path) }
    let tapsMouse = (t.eventsOfInterest & mouseMask) != 0 || t.eventsOfInterest == ~UInt64(0)
    print("tap=\(t.eventTapID) pid=\(t.tappingProcess) proc=\(name)")
    print("  enabled=\(t.enabled) options=\(t.options.rawValue == 1 ? "listen-only" : "FILTER") tapsMouseClicks=\(tapsMouse)")
    print("  min=\(t.minUsecLatency)us avg=\(t.avgUsecLatency)us max=\(t.maxUsecLatency)us")
}
```

### Reproduction steps

1. Have FluidVoice 1.6.9 as a login item, default settings, no mouse-button shortcut configured.
2. An environment with regular audio device-list changes (Bluetooth headphones, Continuity microphone) appears to accelerate it.
3. Boot and use the machine normally.
4. Within a few minutes, trackpad/mouse clicks stop registering system-wide; cursor movement and keyboard are unaffected.
5. `pkill -x FluidVoice` - clicks return immediately.

### Expected behavior

Mouse clicks are never routed through FluidVoice unless a mouse-button shortcut is actually configured - and even then, via a listen-only tap (or a filter tap narrowly scoped to the configured button) so a wedged app can never take out clicking system-wide.

### Actual behavior

Two filter-mode taps intercept all left clicks regardless of configuration. When the tap wedges, clicks queue behind it and die; the disabled-by-timeout callback does not recover it, and the machine is unusable without a keyboard-driven force quit.

### Suggested fixes

- Use `kCGEventTapOptionListenOnly` for mouse events unless a mouse-button shortcut requires suppression, and only include mouse events in `eventsOfInterest` when one is configured.
- Ensure the `kCGEventTapDisabledByTimeout` handler runs on a run loop that cannot be starved by the condition that wedged the tap (see #657, where only the 30-second health check recovers it).

### App Version

1.6.9 (build 20)

### macOS Version

macOS 26.6.2 (25G83)

### Architecture

Apple Silicon

### Logs or crash report

Startup completes normally at 16:28:56; the log is then silent for the whole wedge window until device-list churn at 16:35:30:

```
[16:28:56.618] [DEBUG] [ContentView] Initial hotkey manager health check: true
[16:35:30.568] [INFO] [ASRService] Device list changed - checking for new/removed devices
[16:35:30.596] [DEBUG] [ASRService] Current input devices: iPhone Microphone, MacBook Pro Microphone
[16:35:32.421] [INFO] [ASRService] Device list changed - checking for new/removed devices
[16:35:36.154] [INFO] [ASRService] Device list changed - checking for new/removed devices
```

Full `Fluid.log`, the `sample` output, and the complete tap listing available on request.

### LINKED ISSUE #933: [BUG] IDE keeps re-adding ignored input sources
### Describe the bug

First, I think your allowing us to prioritize them is brilliant! I can place them in the order of most preferred, and as preferred devices come online, your system automatically switches to them. I wish all devices that took inputs did this. 

That said, your logic for ignoring devices essentially ruins that experience when using your MacBook Pro with a dock.

1. You are blindly re-adding devices that have already been ignored when they are re-detected.
2. You are prioritizing those newly detected devices in our device list above our preferred devices. 

Specifically, I use a dock with my MacBook. That dock has both microphone and speaker inputs. However, the microphone that I use is my high-end USB microphone and since nothing is physically connected to the audio inputs/outputs on the dock, I told FluidVoice to ignore them entirely.

The issue is every time I undock and re-dock my MacBook Pro, you keep re-detecting those ignored audio inputs on the dock, then moving them above my USB microphone and other devices, and since nothing is connected to them, essentially your product is now dead until I manually go in and again tweak the settings to put things back, something I have to do on *every* *single* *dock/undock* *event* which can happen multiple times a day. This is beyond frustrating!

To me, this breaks down into two problems. The first is if we mark a device as ignored, it should remain ignored whether it is suddenly removed and re-added physically. You clearly already have logic that handles devices being disconnected/reconnected as devices in the list turn gray when disconnected, and black when they're available again.  Just add state saying a device is permanently ignored and therefore should never be selected.

My suspicion is when you mark a device as being ignored, you're not actually ignoring it.  You're 'forgetting' it by physically removing the device from the list meaning you're no longer aware of it.  Thus when the machine re-docks, it doesn't see a formerly-ignored device, but an entirely new one which it happily adds back.

he second issue is how you handle new devices generally.  You are blindly prioritizing them, placing them at the top of the list which I think is the wrong call. I can understand there may be a benefit to that behavior, but make it a choice, not a rule.

This can be done by adding a simple checkbox labeled "Prioritize newly detected devices". When checked, they're added to the top of the list.  When unchecked, they're added benignly to the bottom.

Implementing either of the above would address the problems I'm dealing with. Implementing both would give us the ultimate flexibility with your system.

### Reproduction steps

1. Attach your MacBook to a dock that has audio inputs/outputs on it
2. Attach a separate USB microphone.
3. Prioritize the order as USB Microphone, Internal MacBook Microphone, Dock Inputs
4. Remove/ignore the dock inputs leaving only the other two
5. Undock your MacBook
6. Re-dock your macbook
7. Not only are the ignored devices back in the list, but they're now prioritized at the top.

### Expected behavior

If I ignore a device, the device stays ignored, period. There's already a button to re-add ignored devices. As I said above, I actually think a better approach would be to leave ignored devices in the list, but with some state saying always slash permanently skip them this is better than removing them from the list where you don't have visibility into them, and as I suspect above, you're actually removing them from your known list of devices so when they are re-detected they are treated as new instead of ignored. 

### Actual behavior

Previously ignored devices were suddenly detected and added to the top of the list, affecting my ability to use your product. 

### App Version

1.6.9

### macOS Version

27 (dev beta 8)

### Architecture

Apple Silicon

### Logs or crash report

```text

```

_No response_

altic-dev:
## Description
- Keep ignored microphones suppressed after disconnecting and reconnecting.
- Stop monitoring global mouse clicks when no mouse shortcut is configured, and harden event-tap cleanup and recovery diagnostics.
- Ignore recording shortcuts while macOS is locked.
- Preserve the hidden-Dock preference when reopening an already-running app.

## Type of Change
- [x] 🐞 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 🧹 Chore
- [ ] 📝 Documentation update

Closes #933
Closes #921
Closes #753
Addresses #884
Related to #657 and #789.

#891 is the separate configured mouse-shortcut ownership case and is not fixed by this PR.

## Testing
- [ ] Tested on Intel Mac
- [x] Ran tests locally: 83 targeted integration tests passed
- [x] Built and installed with `sh build_with_FI_incremental.sh`
- [x] Verified locked-screen shortcuts do not start recording
- [x] Verified the event tap excludes mouse events when no mouse shortcut is configured

## Screenshots / Video

## Notes
The event tap remains filter-mode because configured shortcuts must be suppressible. This PR narrows its event mask to the configured mouse-button families and recreates it only when that configuration changes. #657 still needs affected-machine soak testing; this PR adds clearer immediate recovery diagnostics and safer tap lifecycle cleanup without claiming full resolution.

### DIFF
--- a/Sources/Fluid/AppDelegate.swift
+++ b/Sources/Fluid/AppDelegate.swift
@@ -118,7 +118,9 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
             return true
         }
 
-        // Ensure dock-icon reopen always foregrounds FluidVoice.
+        // LaunchServices can restore the bundle's regular activation policy when
+        // reopening a running app, so reapply the user's Dock preference first.
+        self.applyDockVisibilityPolicy()
         sender.activate(ignoringOtherApps: true)
 
         return !self.bringMainWindowToFrontIfPresent()
--- a/Sources/Fluid/ContentView.swift
+++ b/Sources/Fluid/ContentView.swift
@@ -474,8 +474,8 @@ struct ContentView: View {
                 self.handleRewriteShortcutEnabledChange(newValue)
             }
             .onChange(of: self.pasteLastTranscriptionHotkeyShortcut) { _, newValue in
-                // The hotkey manager reads this value live from SettingsStore, so persisting is enough.
                 SettingsStore.shared.pasteLastTranscriptionHotkeyShortcut = newValue
+                self.hotkeyManager?.refreshMouseShortcutMonitoringIfNeeded()
             }
             .onChange(of: self.isPasteLastTranscriptionShortcutEnabled) { newValue in
                 self.handlePasteLastTranscriptionShortcutEnabledChange(newValue)
@@ -484,6 +484,7 @@ struct ContentView: View {
 
     private func handlePasteLastTranscriptionShortcutEnabledChange(_ isEnabled: Bool) {
         SettingsStore.shared.pasteLastTranscriptionShortcutEnabled = isEnabled
+        self.hotkeyManager?.refreshMouseShortcutMonitoringIfNeeded()
         if !isEnabled, self.activeShortcutRecordingTarget == .pasteLast {
             self.clearShortcutRecordingMode()
         }
--- a/Sources/Fluid/Persistence/SettingsStore.swift
+++ b/Sources/Fluid/Persistence/SettingsStore.swift
@@ -1956,10 +1956,7 @@ final class SettingsStore: ObservableObject {
     func reconcileMicrophonePriority(with devices: [AudioDevice.Device]) {
         var entries = self.microphonePriority
         let preferredUID = self.preferredInputDeviceUID
-        let connectedUIDs = Set(devices.map(\.uid))
-        var suppressedUIDs = self.suppressedMicrophoneUIDs
-        suppressedUIDs.formIntersection(connectedUIDs)
-        self.suppressedMicrophoneUIDs = suppressedUIDs
+        let suppressedUIDs = self.suppressedMicrophoneUIDs
 
         if entries.isEmpty,
            let preferredUID,
--- a/Sources/Fluid/Services/GlobalHotkeyManager.swift
+++ b/Sources/Fluid/Services/GlobalHotkeyManager.swift
@@ -231,6 +231,7 @@ final class GlobalHotkeyManager: NSObject {
     private nonisolated(unsafe) var state = HotkeyState()
     private nonisolated(unsafe) var eventTap: CFMachPort?
     private nonisolated(unsafe) var runLoopSource: CFRunLoopSource?
+    private nonisolated(unsafe) var monitoredMouseButtons: Set<Int> = []
     private let asrService: ASRService
     private var primaryShortcuts: [HotkeyShortcut]
     private var promptModeShortcut: HotkeyShortcut
@@ -510,11 +511,17 @@ final class GlobalHotkeyManager: NSObject {
     private func initializeWithDelay() {
         DebugLogger.shared.debug("Starting delayed initialization...", source: "GlobalHotkeyManager")
 
-        self.initializationTask = Task {
-            try? await Task.sleep(nanoseconds: 1_500_000_000) // 1.5 second delay
+        self.initializationTask = Task { [weak self] in
+            do {
+                try await Task.sleep(nanoseconds: 1_500_000_000) // 1.5 second delay
+            } catch {
+                return
+            }
+
+            guard !Task.isCancelled else { return }
 
             await MainActor.run {
-                self.setupGlobalHotkeyWithRetry()
+                self?.setupGlobalHotkeyWithRetry()
             }
         }
     }
@@ -529,7 +536,27 @@ final class GlobalHotkeyManager: NSObject {
 
     func updatePrimaryShortcuts(_ newShortcuts: [HotkeyShortcut]) {
         self.primaryShortcuts = newShortcuts
+        self.refreshMouseShortcutMonitoringIfNeeded()
         DebugLogger.shared.info("Updated transcription hotkeys", source: "GlobalHotkeyManager")
+    }
+
+    func refreshMouseShortcutMonitoringIfNeeded() {
+        guard self.isInitialized else { return }
+
+        let mouseButtons = self.configuredMouseButtons()
+        guard mouseButtons != self.monitoredMouseButtons else { return }
+
+        DebugLogger.shared.info(
+            "Mouse shortcut configuration changed; rebuilding event tap",
+            source: "GlobalHotkeyManager"
+        )
+        if !self.setupGlobalHotkey() {
+            self.isInitialized = false
+            DebugLogger.shared.error(
+                "Failed to rebuild event tap after mouse shortcut change",
+                source: "GlobalHotkeyManager"
+            )
+        }
     }
 
     func updateCommandModeShortcut(_ newShortcut: HotkeyShortcut?) {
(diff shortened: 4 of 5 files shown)


altic-dev:
Superseded by #937, which merged the independent microphone, Dock, and locked-session fixes. The mouse event-tap changes are intentionally excluded and will be handled separately.

PR #939, opened 2026-09-03
Fix system-wide mouse click interception

### LINKED ISSUE #891: [BUG] Left-click not working on opening the app

### LINKED ISSUE #921: [BUG] Trackpad/mouse clicks stop registering system-wide minutes after login - filter-mode mouse event tap wedges (disabled-by-timeout at 311 s) with no mouse shortcut configured

altic-dev:
## Description
Separates keyboard and mouse event taps so normal clicks never pass through a blocking FluidVoice tap unless a mouse shortcut is configured. Mouse activity used to disambiguate modifier-only shortcuts is observed with a listen-only tap, while configured mouse shortcuts use a narrowly scoped filter with matched down/up consumption and independent recovery. Debug test builds now use a separate preferences domain so tests cannot alter the installed app settings.

## Type of Change
- [x] 🐞 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 🧹 Chore
- [ ] 📝 Documentation update

Closes #891
Closes #921

## Testing
- [ ] Tested on Intel Mac
- [x] Ran tests locally: 86 HotkeyShortcutTests passed

## Screenshots / Video

## Notes
Verified the default configuration has one keyboard filter tap and one listen-only mouse observer tap, with no mouse filter tap. A configured mouse-button shortcut is covered by event-mask and click-consumption tests but was not manually exercised.


davemac:
Sorry been busy - still need me to test this @altic-dev ?

Issue #853, opened 2026-08-13
[FEATURE] Drop AI Enhancement model from memory after inactivity


joshgdale:
### Problem

Currently, even if you have the 'Faster first result' off (on AI Enhancement → Fluid Intelligence), once you've used the model, it then sits in memory until the app is closed.

This is obviously perfect if you use dictation super regularly as it keeps the model warm... but for me (and I assume other users), I might dictate a few hundred words, then nothing for the next hour or so. During that period of inactivity, the `fluid-intelligence-mlx` process is running in the background taking ~3GB memory.

### Proposed solution

To have an additional toggle in the Fluid Intelligence Edit Provider settings that allows the app to drops the model from memory after 3-5 mins of inactivity.

I know that for me, I'd rather have the extra memory on hand (or lower resting memory pressure with all my Chrome tabs, lol) in exchange for a slower cold-start at the beginning of a dictation session. Maybe the inactivity window could be user-customised?

Issue #806, opened 2026-08-07
Fluid-1 (AI Enhancement) model never unloads when idle — same wired-memory issue as #548, but for the enhancement model


jeremy-pixelated:
\`fluid-intelligence-mlx\` loads \`fluid-1-nvfp4-mlx\` + the Gemma MTP drafter (~3.5 GB combined) on first dictation and keeps them pinned in wired, non-pageable memory for as long as the app is running — confirmed via \`vm_stat\` (wired memory tracks almost exactly with the model size) and via the process staying resident for hours with 0% CPU at idle. There is no idle-unload; \`unloadCachedRuntime\` only fires on quit or a config change.

This is the same failure mode as #548 ("Dictation model stays pinned in wired memory and never unloads, freezing 8 GB Macs"), which was fixed for the *speech-to-text* model in #619. That fix doesn't cover Fluid-1 — the AI-enhancement model has no equivalent idle-unload path.

#705 raised this for Fluid-1 specifically and was closed as "intention now" / tracked internally, without a public fix or timeline. Filing this as a focused, standalone tracking issue for that specific gap, since #619 already establishes the pattern (unload after N minutes idle, reload on next use) — it seems reasonable to apply the same fix to Fluid-1.

### Environment
- FluidVoice 1.6.7 (18), MLX backend
- macOS (Apple Silicon)
- Model: \`fluid-1-nvfp4-mlx\` + \`gemma-4-E2B-it-qat-assistant-bf16-mlx-mtp\` drafter

### Suggested fix
Apply the same idle-unload timer used for the ASR model (#619) to the Fluid-1 runtime — unload after N minutes of no dictation, reload transparently on next hotkey press.

Issue #705, opened 2026-07-24
[🐞 BUG] "Faster first result" doesn't keep Fluid-1 ready — first dictation after launch stalls ~8.5 s (model loads only on hotkey press)


matejcermak:
### Description

With **"Faster first result" enabled**, the first dictation after every app launch still stalls for **~8.5 s before the enhanced text appears**. The toggle's tooltip says *"Keeps Fluid-1 ready so your first dictation finishes sooner"*, but Fluid-1 is not actually kept ready: the MLX helper (`fluid-intelligence-mlx serve-json`) sits at ~0 MB RSS after app launch, and the model only starts loading when the dictation hotkey is pressed.

### Environment

- FluidVoice 1.6.5 (16), MLX backend, model `fluid-1-nvfp4-mlx` + MTP drafter `gemma-4-E2B-it-qat-assistant-bf16-mlx-mtp` (draft-block-size 6)
- macOS 26.5.2, MacBook Pro M1 Pro, 16 GB
- "Faster first result" (prefix KV cache) **ON**

### Measurements

Cold pipeline in a fresh helper process (`fluid-intelligence-mlx run`, warm OS caches):

```
latency_ms=9307  ttft_ms=8550  generated_tokens=8  tokens_per_second=10.57
```

- `warmup` subcommand alone (model + drafter load): **~4.7 s** — the remaining ~4 s is the first prefill of the ~1300-token dictation prompt + kernel warm-up.
- Warm requests afterwards are great: `totalMs=1336–3914, ttftMs=836–922, tps=25–33` (prefix cache clearly working).

So the ~8.5 s is paid exactly once per app launch — but it lands on the user's *first* dictation. Short first utterances (a few words) feel the entire stall; only utterances longer than ~9 s of speech hide it, because prewarm overlaps with recording.

### Cause (from source)

`prewarmDictation()` is only triggered from `prewarmPrivateAIDictationIfNeeded`, which is called inside `onCaptureStarted` — i.e. at hotkey press, never at app launch:

- https://github.com/altic-dev/FluidVoice/blob/1b070a8e90e4f9cdba6e3ea6f4a4b03352c26729/Sources/Fluid/ContentView.swift#L3063
- https://github.com/altic-dev/FluidVoice/blob/1b070a8e90e4f9cdba6e3ea6f4a4b03352c26729/Sources/Fluid/ContentView.swift#L3692

[...]

Issue #548, opened 2026-07-07
[BUG] Dictation model (~3.4 GB) stays pinned in wired memory and never unloads, freezing 8 GB Macs


cianhammo:
### Describe the bug

On an 8 GB Mac, dictating loads the model (fluid-1-q4_k_m.gguf, 3.43 GB) into wired, non-pageable memory and keeps it there for the entire time the app is open. It is not released between dictations or when idle, only when the app is quit.

This pins roughly 43% of total RAM as unswappable memory. The system is then forced into heavy swap, causing severe lag, beachballing, and hard freezes that require a force-restart.

Because the model is wired for GPU/Neural-Engine inference, this memory does NOT appear against the FluidVoice process in Activity Monitor's normal Memory (RSS) column. It only shows as system-wide "wired" memory. That makes it very hard for users to attribute the slowdown to FluidVoice. In my case it presented as "5-6 GB wired with no visible process using it," which cost two days of misdiagnosis. Measuring wired memory (top / vm_stat) before and after quitting the app is the reliable way to observe it.

### Reproduction steps

1. Use a Mac with 8 GB RAM. Note baseline wired memory: `top -l 1 | grep PhysMem` (mine: ~1.3 GB wired, ~69% free with the app closed).
2. Launch FluidVoice. Wired rises to ~1.9 GB (model not yet loaded).
3. Dictate once.
4. Observe wired memory jump to ~5.8 GB in a single step, at the moment of dictation (the 3.43 GB model + inference overhead being wired).
5. Stop dictating and leave the app idle. Wired stays at ~5.6-5.8 GB indefinitely. It does not drop between dictations.
6. Quit FluidVoice. Wired immediately drops back to ~1.3 GB and free RAM returns to ~69%.

The swing tracks the app in both directions and matches the model size, confirming the model is held in wired memory for the app's lifetime.

### Expected behavior

The model should be released from memory after a period of idle (e.g. N minutes after the last dictation), or there should be a smaller-model option and/or a low-RAM warning. Wired memory should not remain elevated for the entire app session.

### Actual behavior

Wired (non-pageable) memory jumps by ~3.9 GB the instant dictation starts and stays elevated for the entire app session. It is not released when idle or between dictations, only when FluidVoice is quit.

[...]

PR #943, opened 2026-09-03
feat: unload idle models and show load progress in the notch overlay

HEREISCB:
## Description
Once used, Fluid Intelligence (~3 GB helper) and the speech model stay in memory until the app quits. This adds an idle unload, and fixes the user-facing cost of reloading: the hotkey path never preloaded the speech model while recording, so any load happened after the key was released, and the overlay showed nothing while it waited.

What changes:
- **Idle unload** (Speech settings → "Unload Models When Idle", Never/2/5/10/30 min, default 10). A countdown restarts on every model use and releases both runtimes via the existing `unloadCachedRuntime` / `resetTranscriptionProvider` paths. It defers while recording or processing. An "Unload Now" button is included for testing.
- **Speech model preloads during recording on every path.** `ASRService.start()` now kicks off `ensureAsrReady()` in the background. Previously only the on-screen record button did this; the hotkey path loaded at `stop()`.
- **Streaming preview resumes** if the model becomes ready mid-recording. The first chunk transcribes the buffered prefix.
- **Loading state in the notch overlay**, both Standard and Compact layouts: "Keep talking · loading model" with a progress bar, then "Keep talking · transcribing" until the first preview chunk. Loads report no progress, so the bar eases toward 85% over the last measured load time and completes on ready, the same pattern as Safari's page-load bar. Downloads show a real percentage. Reduce Motion drops the easing and shimmer.
- **Fluid Intelligence warms after the speech model**, not alongside it.

Measured on this machine: Parakeet TDT reloads in 0.1–0.5 s, Fluid-1 + drafter in ~1.7 s with a warm disk cache. With the preload, any dictation longer than that hides the reload entirely.

## Type of Change
- [ ] 🐞 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 🧹 Chore
- [ ] 📝 Documentation update

Refs #853 and #806 (idle unload for Fluid Intelligence), #548 (speech model pinned in memory), #705 (first dictation stalls while the model loads). Closes none directly; #853 and #806 are already closed.

## Testing
- [ ] Tested on Intel Mac
- [x] Tested on macOS version: 26.6.2
- [x] Ran tests locally: 358 tests, 0 failures

Manual: set the timeout to 2 min (or press Unload Now), dictate, confirm the log shows "Idle unload: releasing models"; dictate again and confirm "Model not ready at start; loading in background", "Model became ready mid-recording", and the overlay shows the loading bar before the preview text. Tested in the open-source build; the Fluid Intelligence unload path calls the same `unloadCachedRuntime(reason:)` the settings screen already uses on config changes, but I could not observe the closed-source helper's memory drop from this build.

## Screenshots / Video
Video of the loading bar to follow shortly.

## Notes
- Happy to change the default to Never (opt-in) if you'd prefer nothing to change for existing users.
- The only open-source-visible memory win is the speech model. The 3 GB win is the Fluid Intelligence helper, which this can only trigger through the existing bridge seam.
- Whisper models take several seconds to load, which is where the loading bar earns its keep; Parakeet is fast enough that it mostly shows the warm-up phase.

### DIFF
--- a/Sources/Fluid/ContentView.swift
+++ b/Sources/Fluid/ContentView.swift
@@ -3558,7 +3558,6 @@ struct ContentView: View {
                     TranscriptionSoundPlayer.shared.playStartSound()
                 }
                 self.captureRecordingContext()
-                self.prewarmPrivateAIDictationIfNeeded(for: .primary)
                 DebugLogger.shared.benchmark(
                     "APP_BENCH",
                     message: "overlay_phase phase=recording trigger=first_pcm",
@@ -3570,12 +3569,15 @@ struct ContentView: View {
             }
         }
 
-        // Pre-load model in background while recording (avoids 10s freeze on stop)
+        // Pre-load model in background while recording (avoids 10s freeze on stop).
+        // The AI runtime is warmed only after the speech model is ready so the two
+        // loads don't compete for disk and GPU; both finish long before stop.
         Task {
             do {
                 DebugLogger.shared.debug("ContentView: pre-load model task started", source: "ContentView")
                 try await self.asr.ensureAsrReady()
                 DebugLogger.shared.debug("Model pre-loaded during recording", source: "ContentView")
+                self.prewarmPrivateAIDictationIfNeeded(for: .primary)
             } catch {
                 DebugLogger.shared.error("Failed to pre-load model: \(error)", source: "ContentView")
             }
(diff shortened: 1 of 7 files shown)


HEREISCB:
Closing for now; will resubmit with the video evidence.

PR #966, opened 2026-09-11
Fix microphone recovery after audio device changes

altic-dev:
## Description
Rebuild microphone setup after audio-device changes and fix cancellation and playback restoration when recording recovery fails.

## Type of Change
- [x] 🐞 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 🧹 Chore
- [ ] 📝 Documentation update

Stacked on #955.

## Testing
- [ ] Tested on Intel Mac
- [x] Tested on Apple Silicon Mac (automated tests)
- [x] Ran linter locally: `swiftlint --strict --config .swiftlint.yml Sources Tests Package.swift`
- [x] Ran tests locally: 188 signed tests and 37 controller ThreadSanitizer tests passed.

## Screenshots / Video

### DIFF
--- a/Sources/Fluid/Services/ASRService.swift
+++ b/Sources/Fluid/Services/ASRService.swift
@@ -1346,10 +1346,15 @@ final class ASRService: ObservableObject {
     /// cancellation handler also covers cancellation before queue admission.
     private func startCancellableAudioCapture(
         excluding excludedInputUIDs: Set<String>,
-        forcingInputUID: String?
+        forcingInputUID: String?,
+        onInputSnapshotRead: @escaping @MainActor (Int) -> Void
     ) async throws {
         let task = Task {
-            try await self.startConfiguredAudioCapture(excluding: excludedInputUIDs, forcingInputUID: forcingInputUID)
+            try await self.startConfiguredAudioCapture(
+                excluding: excludedInputUIDs,
+                forcingInputUID: forcingInputUID,
+                onInputSnapshotRead: onInputSnapshotRead
+            )
         }
         self.pendingAudioCaptureBackendStart = task
         defer { self.pendingAudioCaptureBackendStart = nil }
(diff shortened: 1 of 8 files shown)


Sources/Fluid/Services/ASRService.swift
```diff
@@ -2441,15 +2569,9 @@ final class ASRService: ObservableObject {
...
-            self.errorTitle = noUsableMicrophone ? "Microphone Unavailable" : "Recording Error"
-            self.errorMessage = errorMessage
-            self.showError = true
-
-            // Post notification for UI to display
-            NotificationCenter.default.post(
-                name: NSNotification.Name("ASRServiceStartFailed"),
-                object: nil,
-                userInfo: ["errorMessage": errorMessage]
+            self.presentAudioCaptureFailure(
```
grohith327, line 2572: Can you post a system notification maybe instead of directly focusing the main app window ? Clicking on it could take users to the main app

Sources/Fluid/Services/DirectCoreAudioInput.swift
```diff
@@ -809,83 +977,62 @@ final nonisolated class DirectCoreAudioLifecycleController: @unchecked Sendable
...
+        try await self.performHardwareOperation {
+            do {
+                guard self.isShutDown == false,
+                      self.snapshotLock.withLock({ self.shutdownRequested == false })
+                else {
+                    throw Self.error("Direct Core Audio lifecycle is shut down.")
+                }
+                let topologyRevision = self.snapshotLock.withLock { self.hardwareTopologyRevision }
+                if let input = self.input, input.isRunning == false,
+                   topologyRevision != self.lastStartedTopologyRevision
```
grohith327, line 989: Do we want to rebuild the pipeline even if the preferred is mic available (and topology changed due to some device coming/going) ? What would be the latency impact of rebuild this ?

altic-dev: There is some weird issue with one user that doesn't work if I don't do the whole build and gets stuck. the current build I have works for now. If this problem happens, I am not sure about the extra delay as I don't have that info or can reproduce this on my side. So still tinkering with this.

Sources/Fluid/Services/ASRService.swift
altic-dev: Yeah, technically this shouldn't even be popping up at all. But the recovery doesn't happen for the same user I mentioned below and this seems to help. If they have notification suppressed - this will go unnoticed and since it's fatal, it made sense to make the app pop up in front so the user is aware of issue - if it happens - seems to be very rare occurrence.

Issue #865, opened 2026-08-15
[FEATURE] Miniamal recording overlay UI option


rjadithyan:
### Problem

The Ui of the recording overlay takes up a lot of space in the screen (The big black box). I like the fact that we can see the text edits in real time, but from using wispr flow for a long time I like their minimal UI better.

### Proposed solution

I'd like you guys to add a toggle option to switch between the current default recording UI and another minimal UI overlay for the recording. Maybe something that shows up under the notch or just like wispr flow style.

altic-dev:
@rjadithyan check out settings a little more and you'll find everything you want / wish :) 

Thanks. 


we have a lot more customization than anyone else out there!! All you want is already there. 

<img width="1044" height="449" alt="Image" src="https://github.com/user-attachments/assets/b3a9044e-420e-40bb-b769-b7fa690e6504" />

rjadithyan:
appreciate it guys, my bad

grohith327:
Thanks @rjadithyan glad we were able to get it resolved

Issue #743, opened 2026-07-31
[FEATURE] Replacement from current app icon in notch to animated voice reactive animation


dxcore35:
### Problem

<img width="1200" height="904" alt="Image" src="https://github.com/user-attachments/assets/2bce8b33-c6f7-4e89-9d3a-8b204c93993c" />

https://github.com/Jakubantalik/thinking-orbs

### Proposed solution

Never understand why i need to see current active icon in FluidVoice notch (we all know what is current active app). It will be more polished to see some reactive animation when speaking!

altic-dev:
if you're using multiple apps on the same screen - it gives you confidence that the text is going to the right app without searching for the caret :) Think about 25K people using it including you ;) But I Love the design tbh - I'll actually consider some of those designs. I like the sphere!

dxcore35:
There are new orbs added in repo above 👆

Issue #103, opened 2026-01-14
[✨ FEATURE]


AdamElitzur:
**Describe the solution you'd like**
It would be really nice if there are ways to change the styles of the pop-up that comes up when dictating. I know when you put it on the bottom, you're able to change the size, but on the top you're not able to. I'd like a smaller one on the top. And if we could attach it all the way to the top, something like what willow voice does really well.

I just switched from Willow Voice and the design is the one thing I loved more about Willow Voice. It looked like a native apple thing with the notch. Other than that I love FluidVoice!

**Additional context**
Looks like that at the top, very slim and not distracting:

<img width="1087" height="227" alt="Image" src="https://github.com/user-attachments/assets/39f99665-90ea-416b-a579-808196c64c95" />

Thanks!

AdamElitzur:
Hey, thanks so much for the response! Just tried notch with a notch 😂, definitely looks much better.

The one thing I'm wondering, I saw it goes below the notch, maybe there's an option to switch it to extend the notch with a little animation? willow did that in a really cool way, so it like grows out the left and right of the notch, kinda like iphone dynamic island animations. That way it actually looks like it's a physical part of the notch, pretty cool.

Yup like that second part, notch sized overlay, right at the very top of the desktop. Even with monitors, it still looks cool to have that notch shape. Thanks so much!

altic-dev:
<img width="372" height="111" alt="Image" src="https://github.com/user-attachments/assets/aa31dcdc-e910-4ed5-b597-deed7b921b8d" />

<img width="905" height="115" alt="Image" src="https://github.com/user-attachments/assets/267bc5de-b845-42be-91f4-c622e1306c5c" />

altic-dev:
Coming up soon :))

AdamElitzur:
Love this, thanks so much!!

PR #975, opened 2026-09-16
feat: premium overlay styles, Companion, and experimental live typing

kxmXX:
## Description

Adds a set of premium recording-overlay styles, a **Companion** overlay style, and an
**experimental (default-off) live-typing** path, and fixes the issues raised in the
first review round.

This is offered as a graft, not as a demand: take what is useful, drop or gate the
rest. It is a large visual PR, so it is reasonable to only take the visual half.

**No ASR change.** Parakeet/FluidAudio, the audio-capture engine and the mic tap are
untouched. Every new visual only consumes the audio level the app already publishes,
and live typing only consumes the partial transcript the app already computes.

Included:

- **Aurora** rebuilt as a three-layer composition (atmosphere / mass / accent) with a
  shared `AuroraComposition`, so the Settings thumbnail and the live overlay draw the
  same geometry. **Wave**, **Minimal** and a circular **Pulse** whose ring is a
  travelling harmonics wave. `OverlayColorTheme` gains a `highlight` role and the
  gradients/spectrum ramp the styles share.
- **Companion** as a new `OverlayVisualStyle` (not a second window, not an
  `OverlaySize`): a chromeless SVG-like character rendered in the overlay, with six
  elemental variants (fire / water / wind / earth / aurora / gothic), accessories,
  a face that follows the lifecycle, listening ripples and glowing eyes while the mic
  is live. It inherits the existing window, positioning and lifecycle code, so it
  disappears with the overlay like every other style.
- **Zoom band widened** to 25 %–300 %. The Companion's base size was doubled, so the
  previous large look is now its 100 % reference, with its own 50 %–300 % band.
- **Notch**: `NotchPresentationPolicy` now carries the visualizer canvas and icon
  policy; new **Ambient Glow (Beta)** presentation (no icon, no label, a wider 88×24
  canvas); bar count adapts to the canvas width instead of being fixed.
- **Live Typing (Experimental, default OFF)** with three levels — `full`,
  `committedChunks`, `finalOnly` — an automatic downgrade ladder, an ownership model
  of prefix + owned range + trailing text, per-application capability probing, and a
  pasteboard fallback. Gated behind `liveTypingExperimental`; no application is
  certified.
- **Settings**: Companion element/size/accessories inside the style picker, motion
  intensity in the Glow *Advanced* group, a live-typing toggle with a status row, and
  a search-index update.

**Review round 1 — what was fixed**

- **Live Typing is now bound to the recording target.** The session uses the exact
  field captured when recording started (`bindRecordingFocus`) instead of resolving
  "whatever is focused now" on the first partial, and it re-checks that exact element
  before **every** write. A partial can no longer land in a field the user moved to.
- **A session that aborts after writing leaves a tombstone**, so a later partial cannot
  restart a session over the same field and the final transcript is not inserted twice.
- **An absent or empty AX subrole is treated as secure**, so an unreadable field is
  never streamed into. The rule is now a pure helper with direct test coverage.
- **Spoken Send is dispatched** when Live Typing already owns the text: the key is
  posted through an empty plan on the normal safety path and the real outcome is
  reported, instead of returning `.inserted` and reporting a failed send.
- **The clipboard is no longer clobbered.** The fallback writes the transcript as a
  transient/auto-generated item and restores the previous contents a moment later
  unless the user changed the clipboard. The transcript is in the dictation history
  either way.
- **`enableStreamingPreview` restored to its upstream default**, so upgrading never
  silently disables the live preview for users who never touched the setting.
- **Overlay panel re-measures** when the visual style or Companion size changes
  (previously it kept the old frame and clipped the new style).
- **Completion flash** is now cleared after its dwell, not before it.
- **Scale slider labels** now show the real 25 %–300 % band.

## Type of Change

- [ ] 🐞 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 🧹 Chore
- [ ] 📝 Documentation update

Follows the accepted design conversations in
[Discussion #949 — "UI needs a rework."](https://github.com/altic-dev/FluidVoice/discussions/949)
and
[Discussion #963 — "Opt-in typed dictation-companion results and captured-target delivery"](https://github.com/altic-dev/FluidVoice/discussions/963).

`#963` is the reason for the live-typing shape in this PR: it argues that a recording
client must validate the *captured* target rather than current focus, and that a changed
target must fail instead of redirecting output. Those two rules are what review round 1
pushed this PR into implementing. There is no dedicated tracking issue for the visual
work, so if the maintainers would rather fold it into the in-progress UI sweep, this PR
can be closed or reduced to the visual components without any hard feelings.

Related earlier feature requests: #103 (better notch animation), #743 (animated
voice-reactive notch), #865 (minimal overlay option).

## Testing

- [x] Tested on Apple Silicon Mac (arm64)
- [ ] Tested on Intel Mac (no Intel machine available)
- [x] Tested on macOS version: 27.0 (26A428); `MACOSX_DEPLOYMENT_TARGET` stays 15.0
- [ ] Ran linter locally: `swiftlint` is **not installed** in this environment, so it was not run
- [ ] Ran formatter locally: `swiftformat` is **not installed** in this environment, so it was not run
- [x] Ran tests locally: `xcodebuild test -project Fluid.xcodeproj -scheme Fluid -destination 'platform=macOS'`
      (with `CODE_SIGNING_ALLOWED=NO`) → **560 tests, 1 failure**, and that failure is
      pre-existing on `main`: `HotkeyShortcutTests.testKeyboardPayloadIgnoresStrayMouseButtonField`
      (`"Q"` vs `"A"`, an AZERTY-layout artefact present in the clean baseline).

New coverage added by this PR: `OverlayPremiumSettingsTests.swift` (~90 cases) and
`LiveTypingSessionTests.swift` (32 cases), both wired into the test target.

**What was actually verified visually.** The Debug bundle cannot be driven by
Accessibility automation here, so every visual claim comes from offscreen
`ImageRenderer` passes over the **production** components (same structs, same
`AuroraComposition`, same `CompanionGeometry`), plus a real Debug build that was
installed and launched. The screenshots below are those renders.

**What was *not* verified:** signed-release behaviour, real streaming
Parakeet-to-editor delivery, live typing against any third-party app, multi-display
placement, and the completion-flash timing on a real machine. Live Typing ships
default-off for exactly that reason and self-disables on any doubt.

## Screenshots / Video

Offscreen renders of the production components. Top to bottom: styles × formats,
Aurora's three layers across levels and themes, the Companion's variants × lifecycle
states, Companion sizes and accessories, the Pulse ring, and the Ambient Glow notch row.

Overlay styles across formats

Multilayer Aurora across levels and themes

Companion variants and lifecycle states

Companion sizes and accessories

Pulse ring across formats

Ambient Glow notch canvas

(The image files live on a separate `pr-975-assets` branch in the fork so they are not
part of this PR's diff.)

## Notes

- **Live Typing is default-off and experimental on purpose.** It is the part of this PR
  most likely to need upstream redesign; the visual half is independent and can be taken
  on its own.
- **Ambient Glow is labelled Beta.** A light spilling under the cutout is bounded by
  DynamicNotchKit's presentation clipping, so the glow lives inside the content rather
  than around the hardware notch.
- `BottomWaveformView` is kept but is no longer referenced.
- Reduce Motion is honoured by every new visual; no new dependency, no second
  `AVAudioEngine`, no new mic tap; installed preferences are only added to, never reset.

<details>
<summary>Commits</summary>

- `4788c2e` feat: premium overlays companion and experimental live typing
- `1870780` fix: address review feedback on live typing, settings and overlay

</details>


kxmXX:
Thanks @greptile-apps and @chatgpt-codex-connector — all 11 findings were legitimate. They are fixed in `1870780` and the template/policy sections are now in the description.

**Live Typing is now bound to the recording target**
- `LiveTypingController` no longer resolves "whatever is focused now" when the first partial arrives. `ContentView.captureRecordingTargetContext()` calls `bindRecordingFocus(...)` with the same `CapturedFocusTarget` the delivery path already stores, and the session only ever uses that exact element.
- `LiveTypingAXTarget.isStillFocused()` re-checks that exact element before **every** write (reusing `TypingService.isExactFocusTargetActive`), so a partial can no longer land in a field the user moved to. A focus change now aborts the session and suppresses the final paste.

**Aborted sessions keep their tombstone**
- `LiveTypingSession.abortedAfterWrite` is set by `markAborted()` whenever the session stops after it has written. The controller keeps that session until final delivery, so a later partial can not start a fresh session over the same field and the final transcript is not inserted a second time.

**Spoken Send is dispatched**
- `typeOutputPlanToActiveFieldAndWait` now runs `dispatchPostInsertionActionOnly(...)` — the normal safety path with an empty plan — when Live Typing owns the text, and maps the result to `.insertedAndActionDispatched` / `.insertedActionSuppressed` / `.inserted`. The key is posted and `deliverSpokenSend` no longer reports a failed send. The fire-and-forget variant now calls its completion with `.inserted` instead of returning silently, so overlay dismissal and the pipeline summary are preserved.

**The clipboard is no longer overwritten**
- `preserveToPasteboard` snapshots the pasteboard, writes the transcript as a `org.nspasteboard.TransientType` / `AutoGeneratedType` item, and restores the previous contents after a short grace period guarded by `changeCount` — so a copy the user made in the meantime is never clobbered. The transcript is in the dictation history regardless.

**Unreadable subrole is treated as secure**
- The rule is now `LiveTypingAXTarget.isSecureSubrole(_:)`, a pure helper that returns `true` for a `nil`, empty or unknown subrole; both `isSecure()` and `capabilities()` use it. Directly covered by tests, including `nil` → final-only.

**Upgrade no longer disables live preview**
- `enableStreamingPreview` is back to the upstream `?? true` default. That was my regression, not an intentional change.

**Panel re-measures on style change**
- `overlayVisualStyle` and `companionScale` now post `OverlaySizeChanged`, which the overlay already observes for a full re-measure and re-anchor, instead of the unobserved `CompanionSettingsChanged` / no notification at all.

**Completion flash keeps its dwell**
- `clearDeliveryCompletion()` moved to after the 300 ms sleep (and after the generation guard), so the completion visual is actually on screen for its dwell instead of for an incidental frame.

**Scale labels**
- Endpoint labels corrected to 25 % and 300 %.

Tests: `xcodebuild test` → 560 tests, 1 failure, and that failure (`HotkeyShortcutTests.testKeyboardPayloadIgnoresStrayMouseButtonField`, "Q" vs "A") is pre-existing on `main` — it is an AZERTY-layout artefact in the clean baseline.

`swiftlint` and `swiftformat` are not installed in my environment, so those two checkboxes are deliberately left unchecked rather than claimed.

altic-dev:
Oh wow these are pretty insane ones! love the themes too!

